PR #14, opened 2023-07-04
Fixed Delete CR error

fliebenberg:
Have changed the eslintrc.json file to fix the Delete "CR" error.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,6 +5,11 @@
   "plugins": ["@typescript-eslint", "prettier"],
   "rules": {
     "no-console": 1, // Means warning
-    "prettier/prettier": 2 // Means error
+    "prettier/prettier": [
+      "error",
+      {
+        "endOfLine": "auto"
+      }
+    ]
   }
 }


EvgeniiaVak:
@fliebenberg am I correct in understanding that this will silence the checker, if there are some windows-like line ends (and before it would not allow it)?

I would vote for another approach - forcing the auto-formatter to always put linux-like line ends into files. The reason for this is: different people might have different OSs, the ones with Windows would auto-format the files to have additional symbols at the ends of the lines, the linux will auto remove them. The whole reason for adding auto-formatting is that we will not have unnecessary changes in the pull requests, that down the road might get very reluctant to quality-check the code that matters.

fliebenberg:
> 

I am happy with another solution. I am on Windows and it was the files that I got from cloning the repo that had the extra CR at the end of each line. Not sure if this is a setting on your editor?
Anyway, happy if you could suggest another solution. I cant compile any code if the error is not dealt with in some way.

fliebenberg:
Here is a discussion about the problem. https://stackoverflow.com/questions/72578467/prettier-write-replaces-lf-by-crlf-on-windows
It looks like I can set my editor to LF rather than CRLF, but that means everyone else using Windows will have to do the same and I am not sure how that will affect the other projects I am working on. I am thinking it is better to have an automatic solution. The stack overflow answer mentions a git option that might be worth exploring.

fliebenberg:
Can someone on a mac confirm if changing this setting causes issues for them? I am reading some other stack overflow answers and it seems it should still work on macs with the new settings. 
The benefit of making the change in the project eslintrc file is that it only applies the changes to this project. So I (and others do not have to change things in our editors just to conform with this project.

EvgeniiaVak:
Although if the error happens during build time pre-commit hook wouldn't help.

fliebenberg:
> I'm on mac, and all the files I see on the main branch have LF ending. Interestingly in #15 I see no line changes. I guess your editor adds CRLF only to new files, and you see the error during the build. I'm not getting any errors. What does the error you are getting say?
> 
> `"endOfLine": "auto"` means it will allow the CRLF - https://prettier.io/docs/en/options.html#end-of-line I think we need "lf" (which is the default). I agree about automating this. I shared the auto-formatting settings here in the repo (in `.vscode/settings.json`) but of course, they work in vs code only 🤔 Maybe adding a pre-commit hook? https://prettier.io/docs/en/precommit.html

It looks like github converts all commits back to 'lf' in any case, so lets just accept the change and see if it causes any issues. From my reading so far it should not be a problem for mac users (as you have experienced).

EvgeniiaVak (approved):

PR #18, opened 2023-07-05
updates README.md

EvgeniiaVak:
This is the first thing people will see in our repo. What else would be good to add here?


fliebenberg (approved):
Happy with this for now. We can always add more later.

Issue #4, opened 2023-06-28
Setup Next.js to build static site (or switch off SSR)


EvgeniiaVak:
It would be cool to be able to simply host files without running a dedicated server for the frontend. Similar to pure frontend dApp architecture [described here](https://docs-babylon.radixdlt.com/main/getting-started-developers/dapp-frontend/building-a-pure-frontend.html)

fliebenberg:
My understanding is that this will be the type of app we are building. We will not need any smart contract functionality for our app. It will simply interact with the existing alphadex smart contracts through the Radix wallet. The website wont be static though, as it will be receiving live exchange data from the alphadex sdk, which is getting data off the Radix ledger
.

EvgeniiaVak:
I assumed next.js had some build-plugin or config for static sites, but it doesn't look like it, the closest thing I could find to what we need for the document reference is this - https://nextjs.org/docs/pages/building-your-application/optimizing/lazy-loading#with-no-ssr

RepServ:
It's the best path to take. Please what progress has been made so far?

EvgeniiaVak:
@RepServ basically we are developing with this in mind, so it's kinda done. What's left (switching off the SSR) does not mean much logic change, but maybe some settings for deployment.

RepServ:
Well done!

PR #19, opened 2023-07-05
Simple alphadex connection to UI

### LINKED ISSUE #30: Simple alphadex connection to UI

SmashingBumpkin:
Change from the generic template to a very simple overview of data from the adex websocket

Displays:
List of pairs
Some info about currently selected pair
Orderbook for currently selected pair

There are some formatting issues where the orderbook extends into/beyond the page footer, but since the design will be completely changed I didn't fix.

closes #30


package.json
```diff
@@ -13,6 +13,7 @@
     "@types/node": "20.3.3",
     "@types/react": "18.2.14",
     "@types/react-dom": "18.2.6",
+    "alphadex-sdk-js": "^0.5.2",
```
fliebenberg, line 16: Can you please update to the latest version of the SDK. I made one change to export the StaticState type. It is needed for some of the changes I propose below.

src/app/page.tsx
```diff
@@ -1,47 +1,53 @@
...
+import { PairsList } from "./PairsList";
+import { createContext, useEffect, useState } from "react";
+import * as adex from "alphadex-sdk-js";
+import { PairInfo } from "./PairInfo";
+
+export const AdexStateContext = createContext(null);
 
 // more components here: https://daisyui.com/components/
 
+adex.init(); //Connect to alphadex websocket
```
fliebenberg, line 12: Move this line to just below all the imports. Although it is not strictly necessary, it is best to initialise the SDK first before using any of the variables.

src/app/page.tsx
```diff
@@ -1,47 +1,53 @@
-import Image from "next/image";
+"use client";
+
+import { PairsList } from "./PairsList";
+import { createContext, useEffect, useState } from "react";
+import * as adex from "alphadex-sdk-js";
+import { PairInfo } from "./PairInfo";
+
+export const AdexStateContext = createContext(null);
```
fliebenberg, line 8: Add this line just below the moved adex.init() line: 
let initial_staticState = new adex.StaticState(adex.clientState.internalState);

This sets up an initial StaticState variable that we can then use in the context and useState variables. If we dont do this, I get a type checking error because clientState is of type ClintState and the variable we get from the stateChanged$ subscription is StaticState. 

Then change the AdexStateContext line to:
export const AdexStateContext = createContext(initial_staticState);
This just sets the intial value of the context variable to a type StaticType which is the same type we will use to update it later.

src/app/page.tsx
```diff
@@ -1,47 +1,53 @@
...
 
 // more components here: https://daisyui.com/components/
 
+adex.init(); //Connect to alphadex websocket
+
 export default function Home() {
+  const [adexState, setAdexState] = useState(adex.clientState);
```
fliebenberg, line 15: change this line to: 
 const [adexState, setAdexState] = useState(initial_staticState);

Similar to above, we are setting the useState variable to a type StaticType so it can be updated later with the stateChanged$ subscription.

fliebenberg (commented):
I have made a couple of comments of suggested changes to fix type checking errors I am getting for this PR. 
It is actually quite messy, but should work for now to get rid of the type checking errors. I am planning to completely refactor the SDK after RCNet 2 drops and will then aim to clean this up significantly. But for now these changes should solve the issue.

EvgeniiaVak:
@SmashingBumpkin could you please rebase this to after #22 version of `main`? This will help me test locally.

SmashingBumpkin:
I've merged with main @EvgeniiaVak, implemented @fliebenberg changes which have removed the type errors in page.tsx. Also corrected the type errors in PairInfo and OrderBook

EvgeniiaVak:
I updated the code to  removed node_modules to start fresh, installed, it compiles all right but then outputs document is not defined error in console, when I open the page:
```
ev@mac website % npm run dev

> dexter-dapp@0.1.0 dev
> next dev

- ready started server on 0.0.0.0:3000, url: http://localhost:3000
- event compiled client and server successfully in 97 ms (20 modules)
- wait compiling...
- event compiled client and server successfully in 81 ms (20 modules)
- wait compiling /page (client and server)...

🌼 daisyUI 3.1.7 https://daisyui.com
╰╮
 ╰─ ✔︎ [ 2 ] themes are enabled. You can add more themes or make your own theme:
      https://daisyui.com/docs/themes

    ❤︎ Support daisyUI: https://opencollective.com/daisyui

- event compiled client and server successfully in 2.3s (1026 modules)
- wait compiling...
- event compiled successfully in 116 ms (487 modules)
Lit is in dev mode. Not recommended for production! See https://lit.dev/msg/dev-mode for more information.
- error node_modules/@radixdlt/radix-dapp-toolkit/dist/radix-dapp-toolkit.mjs (2498:0) @ Uu
- error ReferenceError: document is not defined
    at eval (./src/app/layout.tsx:20:91)
    at (sc_client)/./src/app/layout.tsx (/Users/evgeniia/repos/website/.next/server/app/page.js:2240:1)
    at __webpack_require__ (/Users/evgeniia/repos/website/.next/server/webpack-runtime.js:33:43)
null

```

Maybe deserves another issue outside of this PR.

src/app/page.tsx
```diff
@@ -1,47 +1,55 @@
...
+
+let initial_staticState = new adex.StaticState(adex.clientState.internalState);
+
+export const AdexStateContext = createContext(initial_staticState);
 
 // more components here: https://daisyui.com/components/
 
 export default function Home() {
+  const [adexState, setAdexState] = useState(initial_staticState);
+  const [hydrated, setHydrated] = useState(false);
```
EvgeniiaVak, line 18: Hydration only makes sense when using SSR, in our app we should be able to go without it theoretically. cc @fliebenberg what do you think?

EvgeniiaVak (changes requested):

EvgeniiaVak:
I love the dedicated component files 👍 very clean

src/app/OrderBook.tsx
```diff
@@ -0,0 +1,20 @@
+import { useContext } from "react";
+import { AdexStateContext } from "./page";
+
+export function OrderBook({ orders }) {
```
EvgeniiaVak, line 4: here there is still an error: "Binding element 'orders' implicitly has an 'any' type.ts(7031)"

EvgeniiaVak (changes requested):

src/app/OrderBook.tsx
SmashingBumpkin: Specified the type, shouldn't occur now

SmashingBumpkin:
I'm using vscode but was manually doing like a sucker. Set it up to auto-format now, much better! :)

src/app/page.tsx
SmashingBumpkin: Initially I left this out, but it caused hydration errors on refresh. 

Is SSR feasible given that the client will be selecting pairs, moving graphs around, creating orders etc?

EvgeniiaVak: I mean SSR (and therefore hydration) only makes sense if we have a server, we won't. I'll try to find a solution with #4

EvgeniiaVak: Although SSG (static site generation) is also a kind of prerendering... and maybe I'm wrong here about hydration 🤔

EvgeniiaVak: Ok, I consulted an independent React developer (pro level): this setup should be ok, if `initial_staticState` will always be the same (otherwise initial state should be `null` and all initialization should be done inside `useEffect` (so that it would always execute client-side (i.e. not during SSG). If not it will give us some errors in the future when it changes. Given that this state contains tokens and pairs, I bet it will change when more tokens appear. cc @fliebenberg

EvgeniiaVak:
closes #30

src/app/page.tsx
EvgeniiaVak: Not even errors - undefined behavior. 
The issue stems from how React works with hydration. During hydration, React needs to reconcile the server-rendered HTML with the client-side state (they need to be identical, if they are not identical in production it's undefined behavior, and the page may start to glitch). 
`useEffect` runs after the render is committed to the screen, meaning it runs after the initial render and after every update. But if you put to console `console.log("initial_staticState", initial_staticState);` just below the initialization (and before the rendering `Home()` function), you can see there are already `tokensMap` and `pairsMap`. I suspect they might change in the future (without changing the library version). But I don't know for sure.

fliebenberg: The initial static state should not change (it is specifically a static state that does not change if things change on AlphaDEX. The state is only updated when a new static state is assigned in the subscription that happens in the useEffect. So I think it should be fine.
Note that all this might change after I refactor the SDK, but we will deal with that at that time.

EvgeniiaVak (approved):

PR #61, opened 2023-08-08
Making orders with wallet v0.3 

EvgeniiaVak:
closes #53 
closes #35

PR #62, opened 2023-08-08
do not merge

EvgeniiaVak:
This is just for comparing the git history, #61 can be merged automatically instead, the conflicts are resolved in the same way here and in #61 the only difference is that the update from main here was with squash and merge, and in #61 with simple merge.


EvgeniiaVak:
closing this one, the real PR is #61

PR #68, opened 2023-08-11
bug fix: fetchBalances is not called when the user makes transactions

### LINKED ISSUE #64: Updates about account balances are not received
I requested some test tokens, but the 
Radix balance: 0
field did not update until I changed the selected pair.

Maybe we need to subscribe to some wallet events somewhere?

EvgeniiaVak:
closes #64

fetchBalances will be called every time when adexState value changes. This should happen rarely.

### DIFF
--- a/src/app/OrderButton.tsx
+++ b/src/app/OrderButton.tsx
@@ -460,15 +460,6 @@ export function OrderButton() {
 
   return (
     <div>
-      <div className="text-sm">
-        <div>
-          {adexState.currentPairInfo.token1.name} balance: {token1Balance}
-        </div>
-        <div>
-          {adexState.currentPairInfo.token2.name} balance: {token2Balance}
-        </div>
-      </div>
-
       <div className="tabs">
         <a
           className={activeTypeTabClass(adex.OrderType.MARKET)}
--- a/src/app/OrderButton.tsx
+++ b/src/app/OrderButton.tsx
@@ -95,21 +95,16 @@ export function OrderButton() {
       account &&
       gatewayApi
     ) {
+      // TODO: Move all logic to separate pure TS class/classes/file/files.
+      // Context providers will recieve only the data necessary to render the page.
+      // This will allow cleaner code and better incapsulation.
       fetchBalances(
         adexState.currentPairInfo.token1.address,
         adexState.currentPairInfo.token2.address,
         account
       );
     }
-  }, [
-    adexState.currentPairInfo.token1.address,
-    adexState.currentPairInfo.token2.address,
-    connected,
-    accounts,
-    gatewayApi,
-    fetchBalances,
-    wallet,
-  ]);
+  }, [adexState, connected, accounts, gatewayApi, fetchBalances, wallet]);
 
   //updates accounts
   useEffect(() => {
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,5 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+
+/logs/
--- a/src/app/Footer.tsx
+++ b/src/app/Footer.tsx
@@ -2,52 +2,8 @@ import "react";
 
 export function Footer() {
   return (
-    <footer className="footer p-10 bg-base-300 text-base-content">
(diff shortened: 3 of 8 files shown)


src/app/OrderButton.tsx
```diff
@@ -95,21 +95,16 @@ export function OrderButton() {
       account &&
       gatewayApi
     ) {
+      // TODO: Move all logic to separate pure TS class/classes/file/files.
+      // Context providers will recieve only the data necessary to render the page.
+      // This will allow cleaner code and better incapsulation.
```
EvgeniiaVak, line 100: Perhaps we could do the trick with redux toolkit (it's not that horrible as the good old vanilla redux) 🤔

EvgeniiaVak: The idea is to prevent errors like that from happening again.

src/app/OrderButton.tsx
```diff
@@ -95,21 +95,16 @@ export function OrderButton() {
...
       fetchBalances(
         adexState.currentPairInfo.token1.address,
         adexState.currentPairInfo.token2.address,
         account
       );
     }
-  }, [
-    adexState.currentPairInfo.token1.address,
-    adexState.currentPairInfo.token2.address,
```
EvgeniiaVak, line 106: Why this fixes the bug?
These addresses don't change when we make a transaction. `fetchBalances` doesn't get called every time when `adexState` is updated, only when we change addresses. So [the subscription event](https://github.com/DeXter-on-Radix/website/blob/fix-balance-update/src/app/layout.tsx#L81) gets ignored.

src/app/OrderButton.tsx
EvgeniiaVak: @fliebenberg @SmashingBumpkin @maelsar are you OK if I refactor the webpage to use https://redux-toolkit.js.org/?

Yes, I was previously opposed to using redux, but since we have so many interconnected state variables I see the point.

# The rationale for refactoring
I just discovered another [similar bug ](https://github.com/DeXter-on-Radix/website/issues/69) having the same origins: Component implementation is mixed with business logic, and so the  business logic contains errors that are hard to investigate/debug/unit-test.

EvgeniiaVak: Alternatively, we could use [useReducer hook](https://react.dev/learn/extracting-state-logic-into-a-reducer) 🤔  I'll experiment with that and will try to fix all the state-related bugs in this PR, converting to draft.

maelsar:
I would also need to update the component I am working on (AccountHistory) when a user makes a transaction.  This should also refresh after the Limit order has been triggered and when a cancel order has been called.

src/app/OrderButton.tsx
EvgeniiaVak: I've experimented a little bit with the built-in useReducer and while it is plausible, it adds almost as much boilerplate as redux-toolkit. Other reasons I now prefer react-toolkit for the website:
- separating API calls from the view will make the code more readable and we will be able to write unit tests
- single state for all app (it appears many components reuse the same updates)
- it will be harder for us to create state-related bugs (redux uses immer.js)
- debugging is easier with the [chrome extension](https://chrome.google.com/webstore/detail/redux-devtools/lmhkpmbekcpmknklioeibfkpmmfibljd)

My hope is to make our lives easier in the long run, apologies for possible code conflicts. Here is the branch I'm doing it in [refactor-redux-toolkit](https://github.com/DeXter-on-Radix/website/tree/refactor-redux-toolkit) - all the components will have dedicated slices that will be responsible for data fetching and converting to values for the views, and the views themselves will just render whatever values are in the state. 

Let me know what you think, if you don't want to work with redux in the future, I can stop and go back to fixing OrderButton state as it is. @SmashingBumpkin @maelsar @fliebenberg

SmashingBumpkin:
Just had a quick look at this and it seems either incorrectly named or not ready for pushing to main?

EvgeniiaVak:
refactoring to redux instead in [refactor-redux-toolkit](https://github.com/DeXter-on-Radix/website/tree/refactor-redux-toolkit) branch

PR #84, opened 2023-08-25
Added utility function to display amounts

### LINKED ISSUE #57: Many digits numbers shorthand
The poll of how to write the huge and tiny numbers in the order-book table resulted in K for 1000s and M for 1000000s - https://t.me/dexter_discussion/4096

There might be some problems with that because it does not solve the very tiny numbers (or might be unintuitive if we use small letters). 

Need to research what numbers there will be in the table, test how they look in the UI, see if we can have any problems and maybe implement another solution.

fliebenberg:
fixes #57 

I have created a new function to display amount given a specified number of digits.
I included quite a few examples in the tests so you can see what it will do in different cases, as it might take too long to try and check the logic. 
It took a bit longer than I thought to write it, but it gets quite tricky if you want the flexibility to specify the number of digits and whether you have a thousands Separator or not.

Branched it off main, so should be quite easy to review and merge.


EvgeniiaVak:
A thing of beauty! I love how the tests are organized 👍 

@fliebenberg Could you also please fix the OrderBook component? It shows a bunch of NaNs for some reason.
Screenshot 2023-08-25 at 16 43 34

fliebenberg:
> 

I would rather not make the changes to the orderbook component as part of this PR. There are quite a few changes required which includes changing away from the no-longer available vars like maxDigitsToken1 etc. I think it is better if the person who built the orderbook makes all the required changes to get it working with the new adex version and then implement the displayAmount function instead of the currently used displayPositiveNumber.

EvgeniiaVak:
@fliebenberg I think you have an outdated orderbook version, I already updated it to the new API and it's in the main branch with the new `AMOUNT_MAX_DECIMALS` and `priceMaxDecimals` 
The `displayPositiveNumber ` is probably indeed broken in `new-amount-display` branch 🤷  or just updating the branch will help . In the main branch the orderbook works fine (although has too many zeros, this why we started the conversation in discord in the first place).
Screenshot 2023-08-28 at 15 55 35

src/app/components/OrderBook.tsx
```diff
@@ -18,9 +18,9 @@ function OrderBookRow(props: OrderBookRowProps) {
     typeof total !== "undefined" &&
     typeof maxTotal !== "undefined"
   ) {
-    const priceString = utils.displayPositiveNumber(price, priceMaxDecimals);
-    const sizeString = utils.displayPositiveNumber(size, AMOUNT_MAX_DECIMALS);
-    const totalString = utils.displayPositiveNumber(total, AMOUNT_MAX_DECIMALS);
+    const priceString = utils.displayAmount(price, AMOUNT_MAX_DECIMALS);
```
EvgeniiaVak, line 21:
```suggestion
    const priceString = utils.displayAmount(price, priceMaxDecimals);
```

src/app/page.tsx
```diff
@@ -1,6 +1,7 @@
 "use client";
 
 import { PairInfo } from "./components/PairInfo";
+import * as utils from "./utils";
```
EvgeniiaVak, line 4: Do we need it here? Shows up as unused in the IDE.

EvgeniiaVak (changes requested):

EvgeniiaVak:
Ah, this is the reason for the constant:
Screenshot 2023-08-31 at 16 19 19
Ok 👍

Issue #36, opened 2023-07-13
Deploy test version


EvgeniiaVak:
This issue is not about the real user-facing deployment (that is blocked by #8 ). This is about deploying a version of the website that our community can interact with during the development. This website should be connecting to the test network.

### pre-requisites

- [x] #27
- [x] #20 

### TODO
- [x] set up CI to build release files - kinda done, but not really needed for deploying the test version #87 
- [ ] ~~hide testnet version from search engines~~ - don't need to do much, because we "use client" in the root currently 😆 it pretty much throws SEO out of the window already 
- [ ] ~~(maybe) deploy to vercel~~ (paid-only for organizations)
- [x] deploy to IPFS 
- [ ] deploy to GitHub

---

Things to consider:
- Maybe we'll need a separate `development` branch, where we push all our usual pull requests, and it is automatically deployed to the test, but maybe having `main` branch directly is ok.
- We will probably need some kind of environment variables for separating testing deployment and real deployment in the future

fliebenberg:
For now it is probably easier to just have a main branch, since everything is still "development". We can look at creating a development branch when we get close to launching the app on the mainnet. That way all PRs go into development and we do pulls into main on either a set timeframe or on feature deployments. I dont have any experience working on repos of this nature. I have mostly worked on my own, which makes things much simpler. So I am keen to hear how you think we should deal with different branches.

EvgeniiaVak:
I agree, for now only `main` branch is good enough.

fliebenberg:
Are you happy for me to close this issue for now? We can raise this again closer to deployment? I think we should have a bigger discussion about the shape of the repo once we get closer to deployment. For now we just work in feature-development branches and the main branch.

fliebenberg:
One suggestion is that we start using a consistent naming convention for branches. When it is a feature we add "feature:" to the front of the branch name and if it is a fix, we add: "fix: " to the front of the branch name. It just makes it a bit clearer when looking through the branches. Happy for other suggestions to achieve the same thing.

EvgeniiaVak:
[...]

PR #87, opened 2023-08-28
add release workflow

EvgeniiaVak:
refs #36

Now think that this probably can be postponed, because it will be more useful for later - when we are worried about breaking stuff and have a separate test and prod versions. So I'll just leave this PR hanging in draft for now.

By test version I mean a version of the website which works only with testnet and can be tested by a few volunteers from the community, who are ready to encounter and report bugs, and that version should be updated as soon as we merge something to main.

The current `release.yml` here is somewhat different  - it makes it possible to build directly GitHub CI by launching manually in GitHub UI: `Actions` -> `Release`. The result is `build.zip` which contains what we usually get in `out` folder when we run `npm run build`. 

The idea is to use releases for later deployment when we are ready to share a bunch of updates together with a wider community (and currently it's to test how it easy or not it would be to deploy in this format). It would also be helpful if we decide to distribute the code for people to deploy themselves.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,53 @@
+name: Release
+
+# manual release
+on:
+  workflow_dispatch:
+
+jobs:
+  release:
+    permissions:
+      contents: write
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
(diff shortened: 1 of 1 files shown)


EvgeniiaVak:
deploying mainnet on netlify with build also running there

PR #89, opened 2023-08-31
Redux-accounthistory

maelsar:
First of all I apologize for the delays.  Bare with me as I am new to redux.
This PR contains the account history features.
Core functionality is ready for testing while enhancements will be in smaller PRs.
- purposely commented out a section inside the accountHistory.tsx.  They will be for future enhancements and I didn't want the testers think its a bug.
- Working CancelOrder
- the history updates when i transaction has been made (both cancel and any new orders)
- Still has a placeholder for Order Price (discussed with Fred) for the; price of how much was the input from the user.
- added a few utils used for display chart.
- shows no message if wallet is connected but doesn't have history.


src/app/components/DisplayTable.tsx
```diff
@@ -0,0 +1,255 @@
...
+  const dispatch = useAppDispatch();
+
+  const handleCancelOrder = (orderId: number, pairAddress: string) => {
+    dispatch(cancelOrder({ orderId, pairAddress }));
+  };
+
+  const filteredData = React.useMemo(
+    () => getFilteredData(data, selectedTable),
+    [data, selectedTable]
+  );
```
EvgeniiaVak, line 73: All the logic should be in redux slices, every time you use `useMemo`, `useState`, `useCallback` it's a signal that maybe what's inside should be moved to a slice, you can create new slices if needed. The .tsx components should be as dumb as possible - only caring about how the data looks, not what data to show (so all filters and such are also flags for moving to slices).

EvgeniiaVak (changes requested):

maelsar:
Will do the changes

maelsar:
Requested changes have been applied.  Cleaned up and organized the slice.
History and Cancel Order still working as intended.

src/app/redux/accountHistorySlice.ts
```diff
@@ -1,14 +1,110 @@
...
+// TYPES AND INTERFACES
+export enum Tables {
+  OPEN_ORDERS = "Open Orders",
+  ORDER_HISTORY = "Order History",
+  TRADE_HISTORY = "Trade History",
+}
 export interface AccountHistoryState {
   trades: adex.Trade[];
+  orderHistory: adex.OrderReceipt[];
+  loading: boolean;
```
EvgeniiaVak, line 21: Is it used somewhere? I only see that it's set and updated, maybe it's something left over from the previous versions and is not needed anymore?

maelsar: loading: boolean;  -> yes this is currently unused.  I kept it for now because im thinking if we might implement in the future for a transaction pop up loading block (for both cancel order and make an order)  or a loader we might be able to use that.
However, if you find it unnecessary then I would be happy to remove it from the code.  Let me know.

EvgeniiaVak: Yes, let's remove and add when/if we use it.

maelsar:
• Removed unused states; error and loading.
• other request about Warning should also be resolved.

EvgeniiaVak (changes requested):
Please update to the latest main branch and resolve the conflicts.

maelsar:
Updated to upstream main.  Fixed conflicts

EvgeniiaVak:
@maelsar the file diff looks strange now, it includes changes that are already in main 🤔 I think it's probably because you used rebase. Probably nothing bad will happen if I merge, but just to have a cleaner diff please use simple merge (without squashing) in the future. Discussion about it here: https://discord.com/channels/1125689933735145503/1126739703559897209

maelsar:
I apologize and I will keep that in mind in the future.

EvgeniiaVak (approved):

EvgeniiaVak:
ok, I cannot properly test currently (because of all the versions of APIs and wallets mismatch), but the code looks all right...  merging 🤞

PR #90, opened 2023-09-01
chart clipping fix

maelsar:
Tiny fix.
Problem: time scale doesn't appear on the chart. After some testing I found that lightweight charts keeps applying overflow: hidden on the div of the charts which causes the clipping.

This small small piece of code seems to have fixed the problem.

Now chart shows the timescale.


SmashingBumpkin (commented):
This works for me across firefox, edge and chrome. @maelsar Did you test on safari? If so let me know and I'll approve, otherwise I'll let someone test safari first.

maelsar:
I am unable to test this on safari.

SmashingBumpkin:
@EvgeniiaVak Could you test on safari and approve?

EvgeniiaVak (approved):
works in safari as well 👍

PR #91, opened 2023-09-01
fixes amount undefined

EvgeniiaVak:


SmashingBumpkin:
This is an improvement, though still doesn't return the balances on first load, you have to change pairs first to rejig it. Also, I noticed when changing pairs quickly, sometimes the state balance doesn't get updated and an incorrect balance is shown.

I can raise a new issue if that's beyond the intended scope of this PR?

EvgeniiaVak:
@SmashingBumpkin 

> still doesn't return the balances on first load

this is weird, it loads fine for me, did you clean the browser cache after switching to the branch (hard refresh)? 

> when changing pairs quickly, sometimes the state balance doesn't get updated and an incorrect balance is shown

this is bad, we definitely need to fix, although I tried to reproduce and couldn't - either I'm too slow or it's also something about browser cache

yes, of course you can create an issue, but probably we can fix in this very branch, or merge this one and create a separate

SmashingBumpkin:
> @SmashingBumpkin
> 
> > still doesn't return the balances on first load
> 
> this is weird, it loads fine for me, did you clean the browser cache after switching to the branch (hard refresh)?

Sorry I'm unable to test anything since my wallet app updated, and before my brain kicked in this morning I signed out of the wallet to try to reset it.

abdulhaseeb13mar (approved):

abdulhaseeb13mar:
@EvgeniiaVak @SmashingBumpkin please resolve the conflict so we can merge it

abdulhaseeb13mar (approved):

PR #93, opened 2023-09-02
config: absolute path imports

abdulhaseeb13mar:
now you can use your import paths like this i.e 
import Navbar from "components/NavBar";
you can directly refer everything inside src/app folder now.

### DIFF
--- a/src/app/components/OrderInput.tsx
+++ b/src/app/components/OrderInput.tsx
@@ -365,7 +365,7 @@ function SubmitButton() {
   return (
     <div className="flex flex-col">
       <button
-        className="flex-1 btn btn-primary"
+        className="flex-1 btn btn-accent"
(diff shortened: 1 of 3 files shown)


abdulhaseeb13mar:
@EvgeniiaVak please review and merge these changes.

abdulhaseeb13mar:
@EvgeniiaVak i have done the change that you requested

abdulhaseeb13mar:
> These changes are good! I especially like the colors. One tiny request is to update the `btn-primary` to `btn-accent` in the `OrderInput.tsx`. It might cause a tiny code conflict later, but I think it would be worth it, look how juicy the button is in this color (and without it it's not very visible with the new colors):
> 
> Screenshot 2023-09-04 at 17 24 40

@EvgeniiaVak i have done the change that you requested

EvgeniiaVak (approved):

Issue #72, opened 2023-08-15
Order button updates


SmashingBumpkin:
I'm just opening this issue to consolidate all the feature adjustments needed for the order button. This consolidates previous issues/code todos.

New features:

- [x] Check user has funds to cover position size, show alert message if not (https://github.com/DeXter-on-Radix/website/issues/50)
- [ ] Check user is not using 100% of available xrd (https://github.com/DeXter-on-Radix/website/issues/50)

Feature adjustments/improvements:
- [x] Change price on a limit order is 5% below or above best buy/sell alert to be less intrusive (eg toast) (https://github.com/DeXter-on-Radix/website/issues/50)
- [x] Change slippage greater than 5% alert to be less intrusive (eg a toast) (https://github.com/DeXter-on-Radix/website/issues/50)
- [x] Adjust % button behaviour to put in the requested amount even if there is not liquidity
- [ ] Improve wording and information given in trade "description"
- [x] Quickly typing in a percentage position size on the "Buy" side sends a call to fetch balances too quickly. Might need debouncing

Other updates:
- [x] Automatic updating of balances (will be addressed in this issue: https://github.com/DeXter-on-Radix/website/pull/68)
- [x] Getting best buy/best sell prices (should eventually be addressed by this issue: https://github.com/DeXter-on-Radix/website/pull/68)
- [x] Implement proper design/styling

EvgeniiaVak:
fixed by #117

PR #105, opened 2023-09-09
Order button improvements

SmashingBumpkin:
refs #72 
refs #79 

This PR makes a few changes to the order button.

- Displays appropriate warning messages when "extreme" numbers are input
- Uses utils to display cleaner numbers
- Adds more code comments
- Catches more errors before making calls (eg adex quote checks that a position size is input, gateway checks that a wallet is connected)
- Makes eslint complain about unused import and fixes eslint img warnings (with the aim of making further PR reviews easier)

I wanted to get this in before I dissappear for a couple of weeks. Please feel free to step in and take the lead on this


EvgeniiaVak (approved):
I think this one is good for merge, but since I also added a few things, it would be good if somebody else also took a look before merging.

abdulhaseeb13mar (approved):
seems good to me

Issue #76, opened 2023-08-21
Decide what to use: `.` or `,` in numbers and test input fields


EvgeniiaVak:

fliebenberg:
I suggest we read the current Locale number format and get the decimal and thousands separator from there and store them as variables to be used in displayAmount.

fliebenberg:
OK, we have decided to not implement the locale specific separators for the MVP, so I suggest we just go for "." for decimal separator and " " for thousands separator. Any alternative suggestions?
BTW, these are the defaults in the displayAmount function at the moment.

PR #112, opened 2023-09-12
Read locale separators

fliebenberg:
Addresses issue #76
I added a utility function to read the locale specific separators to use when displaying amounts.
I added this function to run when the app initialises and store the separator values in a ui state.
The displayAmount function then uses these separtors by default.

### DIFF
--- a/src/app/redux/UiSlice.ts
+++ b/src/app/redux/UiSlice.ts
@@ -1,0 +1,26 @@
+import { createSlice, PayloadAction } from "@reduxjs/toolkit";
+
+export interface UiState {
+  decimalSeparator: string;
+  thousandsSeparator: string;
+}
+
+const initialState: UiState = {
+  decimalSeparator: ".",
+  thousandsSeparator: " ",
+};
+
+export const uiSlice = createSlice({
+  name: "ui",
+  initialState,
+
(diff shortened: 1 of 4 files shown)


fliebenberg:
@EvgeniiaVak could you just check that I implemented the redux state stuff correctly. It has been a while since I worked with redux and I did not want to spend too much time to learn it again.
Also, I cannot run the utils tests at t he moment as it tells me it cannot find the redux/store file (you should be able to reproduce the problem by trying to run tests in this branch). I suspect it has something to do with the jest setup in that it uses a different path to run the tests and therefore cannot find the path as specified, but it might be something completely different. Would appreciate if you can see if you can fix the problem. I can then try and incorporate some tests for different locales to see if the solution works.

EvgeniiaVak:
@fliebenberg I think we can just hardcode the separators and such in the `utils` themselves, and not import anything redux to `utils` at all. Maybe it's good to have for user settings but we did not discuss if it's needed at all and I don't think we need to spend time on it now, as different user preferences are not in the MVP.

fliebenberg:
Cool. We can always revisit it at a later stage. It is not a big change to make later again. 
So I will close this PR then.

PR #113, opened 2023-09-13
charts enhancement

### LINKED ISSUE #74: Price Chart updates
- [x] x-axis time scale rendering
- [x] on hover info (above the chart?)
- [x] more visually integrated candle period selector

### LINKED ISSUE #97: Price Chart styling

maelsar:
fixes #97 
fixes #74 

NOTE: I still have further improvements in the pipeline (add more chart options, formatting, layout), but this should be stable enough for the MVP and for testing by the community in the meantime.
#74 

In the meantime, if its alright, I will work on the Price Info component so we can get most of the components ready for the MVP.


src/app/redux/priceChartSlice.ts
```diff
@@ -54,6 +103,28 @@ function convertAlphaDEXData(data: adex.Candle[]): OHLCVData[] {
...
+        setLegendPercChange({
+          currentOpen: latestOHLCVData.open,
+          currentClose: latestOHLCVData.close,
+        })
+      );
+      dispatch(setLegendCurrentVolume(latestOHLCVData.value));
+    }
+  }
+);
+
```
EvgeniiaVak, line 127: I don't think we need to create a separate thunk for that, we are not doing any network requests here or something like this, all the data comes from adex - we need to update every time adex updates (and this is just not the same as triggering it in `useEffect` (it will most likely work because `data` variable will be updated more often than anything else, but still it's not the most straightforward way).

I suggest to create `updateAdex` in priceChartSlice reducers, accept `action: PayloadAction<adex.StaticState>` as the second param there, and update all the necessary state changes from it, this way we won't miss the updates and we won't need to call `fetch...` in the component (we already subscribe to the changes of adex state).

EvgeniiaVak (changes requested):

EvgeniiaVak:
Screenshot 2023-09-15 at 10 37 26

Maybe making this padding smaller?

maelsar:
> Screenshot 2023-09-15 at 10 37 26
> Maybe making this padding smaller?

that space doesn't have contain any padding. That is coming from lightweight charts.  I am still looking into if there is an option to remove that from the chart for a future PR since I wanted avoid doing something like negative margins.

src/app/redux/priceChartSlice.ts
maelsar: i refactored the code not to use a thunk.  Important to note this is only meant to initialize the legend only when the page is first loaded or move to a different time selector.
Since the crosshair is functional and updates the legend to the candle pointed, I didn't want the legend to update to the current candle data when a new candle is printed.  It might be disruptive if a user is on a 5m timeframe and points to a certain candle.

EvgeniiaVak:
oh, I there probably was a reason for this padding... now on some screen widths the chart jumps on the page (becomes taller0 when there are lots of digits displayed:
Screenshot 2023-09-18 at 11 22 26
Screenshot 2023-09-18 at 11 22 16

but it definitely looks better without the padding when there are few digits 🤔  maybe we can round up the digits? we have functions for that in utils, but I don't know if this is a place where traders expect precision 
@fliebenberg what do you think?

src/app/redux/priceChartSlice.ts
EvgeniiaVak: I don't really get the reason behind `initializeLegend`, why can't we always show what adex provides? Like you did in priceInfoSlice.

fliebenberg:
It should be fine to round these numbers. No one should be interested in more than 8 decimals as that is what AlphaDEX works with.
Also, if we are going to be showing the high/low and other stats in the graph, then we do not need the 24h stats above the graph. It is basically giving the same values, although I am not sure how the Trading View values are calculated.

EvgeniiaVak:
> we do not need the 24h stats above

in the chart we can choose not only 1 day period, it can be something different like 5 minutes, so it's not always duplicating info and I think it's ok to always have the price info also

src/app/components/PriceChart.tsx
```diff
@@ -37,30 +58,45 @@ function PriceChartCanvas(props: PriceChartProps) {
...
-        color: "#eaeff5",
       });
 
-      volumeSeries.setData(clonedData);
+      // VOLUME BARS
+      // MODIFY THEME COLOR HERE
+      volumeSeries.setData(
+        data.map((datum) => ({
+          ...datum,
+          color: datum.close - datum.open <= 0 ? "#ef5350" : "#26a69a",
```
EvgeniiaVak, line 84: These colors should come from css variables. As strings in code they can be represented as `hsl(var(--su))` for green, `hsl(var(--er))` for red.

src/app/components/PriceChart.tsx
```diff
@@ -4,30 +4,51 @@ import {
...
+          textColor: "#DDD",
+        },
+        //MODIFY THEME COLOR HERE
+        grid: {
+          vertLines: { color: "#444" },
+          horzLines: { color: "#444" },
+        },
         timeScale: {
+          //MODIFY THEME COLOR HERE
+          borderColor: "#d3d3d4",
```
EvgeniiaVak, line 51: These colors probably can also come from the theme settings. You can find what colors we currently have at `tailwind.config.js` under `daisyui` objects (only dark theme currently).

Here is the list of color names and variables that you can use - https://daisyui.com/docs/colors/#-2 . Here in the chart settings you can use them as variables (e.g. `hsl(var(--b1))`). In other places where you style an html element in .tsx you can specify them as classes, e.g. for background `bg-base-100`  or for text color `text-base-100`.

src/app/components/PriceChart.tsx
```diff
@@ -70,37 +106,100 @@ function PriceChartCanvas(props: PriceChartProps) {
...
+            <span
+              className={isNegativeOrZero ? "text-red-500" : "text-green-500"}
+            >
+              {candlePrice?.open}
+            </span>
+          </div>
+          <div>
+            High:{" "}
+            <span
+              className={isNegativeOrZero ? "text-red-500" : "text-green-500"}
```
EvgeniiaVak, line 163: Here and below (and you probably can specify `{isNegativeOrZero ? ...}` one time in the enclosing div (the one with the `<div className="flex justify-between`) also please use the theme colors to make page look consistent.

Please also remove `font-roboto` I think it should be set up for the whole app.

EvgeniiaVak (changes requested):

src/app/redux/priceChartSlice.ts
maelsar: @EvgeniiaVak I have tried the approach initialize the data with adex.clientState.currentPairCandlesList (similar to what you mentioned priceInfoSlice). The issue might be with initialization timing, the state is being initialized before the data in adex.clientState.currentPairCandlesList is populated thus making the legends all initially empty when i first load the page.

I purposely wanted to prevent updating the legend when a new candle appears that why i did not go with the approach to update the legend on state change.

maelsar:
I think I updated the code when you saw this that why it no longer showed that large spacing between the legend and chart.

EvgeniiaVak:
> I think I updated the code when you saw this that why it no longer showed that large spacing between the legend and chart.
> 
> 

That's fine, given that [we can shorten them](https://github.com/DeXter-on-Radix/website/pull/113#issuecomment-1722932277) we probably won't have this problem. @maelsar would you like to update this PR with using utils for shorter numbers and other changes I requested above (styling using the theme config, etc.)?

src/app/components/PriceChart.tsx
nguvictor: When I tried to use the hsl(var(--su)) variables it cause the chart to be black (randomly hit or miss), I suspect it was because the library does some extra steps with the string
image

EvgeniiaVak: ok, probably we can import tailwind config object then and take the color from there.

src/app/components/PriceChart.tsx
nguvictor: I removed the font-roboto and swapped colors where I could  in https://github.com/DeXter-on-Radix/website/pull/145

EvgeniiaVak:
these commits are included in #145

Issue #103, opened 2023-09-06
OrderBook styling for dark theme


EvgeniiaVak:

EvgeniiaVak:
fixed by #114

PR #114, opened 2023-09-13
Order Book UI updates

EvgeniiaVak:
refs #103


src/app/components/OrderBook.tsx
```diff
@@ -109,31 +113,36 @@ export function OrderBook() {
...
-        <div className="text-end">
-          Price
-          <br />({token2Symbol})
-        </div>
-        <div className="text-end">
-          Size
-          <br />({token1Symbol})
+    <div className="p-2 text-sx text-primary-content">
+      <div className="sized-columns">
+        <div className="mx-2 my-4 whitespace-nowrap text-secondary-content">
```
abdulhaseeb13mar, line 118: @EvgeniiaVak we do not need Order Book now, it was just placed by me for understanding the placement of the layout component

EvgeniiaVak: I took it from Vlad's figma. Also, I think I like it, but we can remove of course. Did you try with and without it, what looks better to you?

EvgeniiaVak: ok, removing, it's ok without it too. Oh, also I forgot the narrow screen fix, back to draft for now.

abdulhaseeb13mar: i think the order book component is pretty much self explanatory to the users. and besides we are not giving headings to other components. so i would suggest to keep it consistent with other components

abdulhaseeb13mar: > i think the order book component is pretty much self explanatory to the users. and besides we are not giving headings to other components. so i would suggest to keep it consistent with other components

referring to this figma design 
image

EvgeniiaVak:
Atm I don't have much time for making the nice collapsible thing for mobile screens, but at least orderbook is visible now there too.

abdulhaseeb13mar (approved):

abdulhaseeb13mar:
@EvgeniiaVak Looks good to me you can merge now

Issue #121, opened 2023-09-19
Price Info styling


EvgeniiaVak:

nguvictor:
Made an PR to address this issue, looking forward to feedback!


![Image](https://github.com/DeXter-on-Radix/website/assets/3978285/47723c20-b878-4978-b655-b800262c7634)

PR #126, opened 2023-09-22
Price info styling

nguvictor:
Fixes https://github.com/DeXter-on-Radix/website/issues/121
image


src/app/components/PriceInfo.tsx
```diff
@@ -3,63 +3,83 @@ import { useAppSelector } from "../hooks";
...
-  const lastPrice = priceInfo.lastPrice;
-  const change = priceInfo.change24h;
-  const high = priceInfo.high24h;
-  const low = priceInfo.low24h;
-  const volume = priceInfo.value24h;
+  const lastPrice = priceInfo.lastPrice.toFixed(3);
+  const change = priceInfo.change24h.toFixed(2);
+  const high = priceInfo.high24h.toFixed(3);
+  const low = priceInfo.low24h.toFixed(3);
+  const volume = priceInfo.value24h.toFixed(3);
```
EvgeniiaVak, line 10: Maybe using the `displayAmount` function from utils instead of `.toFixed` here? It shortens very large numbers with K and M (like in order book).

src/app/components/PriceInfo.tsx
```diff
@@ -3,63 +3,83 @@ import { useAppSelector } from "../hooks";
...
-        <span className="text-sm">24h Change</span>
+        <span className="text-xs font-bold text-secondary-content uppercase pb-1">
+          24h Change
+        </span>
         <span
           className={
-            isNegativeOrZero ? "text-xs text-red-500" : "text-xs text-green-500"
+            isNegativeOrZero
+              ? "text-sm font-bold text-red-500"
+              : "text-sm font-bold text-green-500"
```
EvgeniiaVak, line 42: Instead of `text-red-500` and `text-green-500` please use colors from the theme (they are specified in `tailwind.config.js`, I believe these should be `text-error` or `text-success` (but maybe `text-error-content` or `text-success-content will look better`... I don't know, please check.

I would even go with what [Vlad defined in figma](https://www.figma.com/file/qushS15h87wjpoXaY6ND4W/DeXter-UI---refresh?node-id=501%3A1107&mode=dev) and dropped the conditional `isNegativeOrZero` coloring (and removed the variable from the state at all) and color only the number in `24HRS CHANGE` column with accent color (`text-accent`).

EvgeniiaVak (changes requested):

src/app/components/PriceInfo.tsx
nguvictor: Nice, didn't know we had a utility class for this.

src/app/components/PriceInfo.tsx
nguvictor: text-error and text-success looked better when I test vs text-error-content text-success-content

nguvictor:
Updated with suggestions
image

EvgeniiaVak (changes requested):
I just checked on https://deploy-preview-126--chipper-sunburst-578cfe.netlify.app/

Screenshot 2023-09-26 at 15 22 33

it says `+1.28%` but the color is red. Something is wrong here. Either the change is actually negative or the color should be green (if we go with the green/red colors changing still, if we go with accent it's always one color).

nguvictor:
@EvgeniiaVak you are correct I made a typo when swapping out the colors. Personally I like the change in color, but this is a very easy tweak if we get consensus in the future.

EvgeniiaVak (approved):

PR #128, opened 2023-09-25
feat: account history component style and code revamp

### LINKED ISSUE #120: Account History styling

abdulhaseeb13mar:
fixes #120 

i have opened this PR, for now it is not responsive for tab and mobile. will be working on it's responsiveness in the coming days


EvgeniiaVak (changes requested):
I tried to have an open order and the order count indicator is drawn on the second row (my window width is 1470px here):

Screenshot 2023-09-26 at 16 27 48

Is it possible to show them on the same row?

abdulhaseeb13mar:
> I tried to have an open order and the order count indicator is drawn on the second row (my window width is 1470px here):
> Is it possible to show them on the same row?

@EvgeniiaVak i have fixed it, you can check now

EvgeniiaVak (changes requested):
Border radii should probably be 0 for both the count indicator and the cancel button on hover to look consistent with the rest of the styling, which is quite rectangular. Strange that they are not absent by default on the button, we have it in the tailwind config for all the buttons.

The `cancel` button (when without hover) looks a little bit out of place - shifted to the right, 🤔 maybe moving the cancel button a little bit (e.g. use negative padding? what can go wrong? 😆 ). Or maybe remove the on hover at all, or only underline on hover? I vote for only underlining, with no bg changes.

Other differences from figma:
1. all headers are uppercase
2. `cancel` is red 
2. the selected tab has accent color

Screenshot 2023-09-26 at 18 55 05
Screenshot 2023-09-26 at 18 57 03

EvgeniiaVak (approved):

nguvictor (approved):

EvgeniiaVak (approved):
@nguvictor reviewed (thanks), so approving this PR.

PR #130, opened 2023-09-26
pair selector and style

### LINKED ISSUE #122: Pair selector styling + shortcut

nguvictor:
Fixes #122 
swap-pair

### DIFF
--- a/src/app/components/PairSelector.tsx
+++ b/src/app/components/PairSelector.tsx
@@ -28,13 +28,13 @@ export function PairSelector() {
       </label>
       <ul
         tabIndex={0}
-        className="dropdown-content z-[1] menu shadow bg-neutral rounded-box w-full !my-0"
+        className="dropdown-content z-[1] menu shadow bg-neutral rounded-box w-full !my-0 !px-0"
       >
-        <li className="text-xs text-left uppercase text-secondary-content px-4">
+        <li className="text-xs text-left uppercase text-secondary-content !pl-4 ">
           Select a pair of tokens:
         </li>
         {pairSelector.pairsList.map((pair, index) => (
-          <li className="font-bold" key={index}>
+          <li className="font-bold !pl-0" key={index}>
             <button
               className="justify-between"
               onClick={() => selectPair(pair.address)}
--- a/src/app/components/PairSelector.tsx
+++ b/src/app/components/PairSelector.tsx
@@ -8,13 +8,10 @@ export function PairSelector() {
     dispatch(selectPairAddress(pairAddress));
   };
   return (
-    <div
-      className="dropdown dropdown-start w-full not-prose"
-      id="pair-selector"
-    >
+    <div className="dropdown dropdown-start w-full" id="pair-selector">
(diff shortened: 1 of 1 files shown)


src/app/components/PairSelector.tsx
```diff
@@ -8,18 +8,42 @@ export function PairSelector() {
     dispatch(selectPairAddress(pairAddress));
   };
   return (
-    <div className="dropdown" id="pair-selector">
-      <label tabIndex={0} className="btn m-1">
-        {pairSelector.name}
+    <div
+      className="dropdown dropdown-start w-full not-prose"
```
EvgeniiaVak, line 12: it would be super cool to try and not specify `not-prose` - we want to be able to have consistent fonts throughout the website

if prose overrides some styles that you want to apply, you can override them once again with `!`, e.g. `!m-0` instead of `m-0`

EvgeniiaVak (changes requested):
Something is off with paddings / margins there, the selector looks a little bit out of position.

Screenshot 2023-09-26 at 16 39 51


Also background color - maybe making it the same as for other components (like in figma)?

src/app/components/PairSelector.tsx
nguvictor: Nice, I'm new to daisyui and tailwind in particular. Didn't know they had an !m-0. I just saw in the css that non-prose was overwriting when I was using my-0.

nguvictor: It looks like the colors are a bit off in the theme. The figma says #191B1D which is base-100 in the theme. But the color for the rest of the background items are #1a1c1e. We might want another issue to line up the colors across the dapp.

image

nguvictor:
It looks like the colors are a bit off in the theme. The figma says #191B1D which is base-100 in the theme. But the color for the rest of the background items are #1a1c1e. We might want another issue to line up the colors across the dapp.

image

src/app/components/PairSelector.tsx
```diff
@@ -8,18 +8,39 @@ export function PairSelector() {
...
         {pairSelector.pairsList.map((pair, index) => (
-          <li key={index}>
-            <button onClick={() => selectPair(pair.address)}>
+          <li className="font-bold !pl-0" key={index}>
+            <button
+              className="justify-between"
+              onClick={() => selectPair(pair.address)}
+            >
               {pair.name}
+              <span>+</span>
```
EvgeniiaVak, line 43: Why is there a `+` sign?

EvgeniiaVak:
I suggest to remove bg color at all and use the default one, and then change the default one for the correct `#191B1D` (either in a separate pr or here)

EvgeniiaVak (approved):

src/app/components/PairSelector.tsx
nguvictor: The figma had a + sign, I could of used a svg icon, but seemed overkill. I suspect this will be a favorite button in the future, but its a placeholder for now.

PR #134, opened 2023-09-27
Navbar update to match figma

### LINKED ISSUE #133: Update Navbar to match Updated Figma
Vlad's UI Refresh
https://www.figma.com/file/qushS15h87wjpoXaY6ND4W/DeXter-UI---refresh?mode=dev

nguvictor:
Fixes #133 

Colors need to be tweaked a bit and some padding polish, but I can do that after the first initial reviews on nav update to align figma.
navbar refresh

Note:
Didn't change the default page.tsx that has the current trade page as I didn't want to disturb existing PR's
Placed some filler pages for Landing, Trade, Markets
Landing: We can remove this link and put it at the root, when everyone is ready
Trade: Future location of the trade page, placeholder for now
Markets: Mock out to match Figma

### DIFF
--- a/src/app/components/NavBar.tsx
+++ b/src/app/components/NavBar.tsx
@@ -1,16 +1,17 @@
 import React from "react";
 import Image from "next/image";
 import { FaRegMoon } from "react-icons/fa6";
+import Link from "next/link";
 
 // TODO: theme switching
 
 export function Navbar() {
   return (
-    <div className="col-span-12 navbar">
-      <div className="navbar-start">
-        <a
+    <div className="col-span-12 navbar p-0 min-h-6">
+      <div className="flex-none px-2 mx-2">
+        <Link
           href="/"
-          className="btn btn-ghost hover:bg-transparent no-underline text-xl p-0"
+          className="btn btn-ghost hover:bg-transparent no-underline text-xl p-0 join-item"
         >
           <Image
             src="/logo_icon.svg"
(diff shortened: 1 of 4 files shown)


EvgeniiaVak:
I think some of this PR (making it a bit more like figma, even though we won't have markets or community tabs yet) can still probably go into the MVP:
- trade tab (even though it's the only one we have so far... we'll just add others later)
- removing DEXTER from Navbar
- hiding the theme button
- fixing paddings

@nguvictor would you like to continue working on it?

nguvictor:
Yep, I'll can finish it up the PR.

Personally, I was debating on leaving the DEXTER, so that any screenshots have the name for people to look up or share. Otherwise there are no places where you can see the DEXTER name brand only logos at the moment.

I'll follow the figma for now, we can tweak later if anything

EvgeniiaVak:
Yeah, we can leave DEXTER as is, it's not like navbar is overpopulated right now.

EvgeniiaVak (changes requested):
DEXTER looks somewhat out of place... maybe it's the underlining on hover (which is different from the rest of the tabs)... it's ok that something is different about that link to home, but idk looks small compared to the trade and market. Maybe making font size larger on hover (with transition and font size tailwind classes)?

Also maybe a bit larger spacing between the logo image and DEXTER

fliebenberg:
I also like the DEXTER when there is space. We can hide it on mobile screen sizes. Agree that it needs to be bigger. I am not sure if we are looking at the same figma files, but the figma files from the linktree links shows the DEXTER a bit bigger, so I would just go with something similar to that.

fliebenberg:
Here is the link to the figma files that is being used for this PR: https://www.figma.com/file/qushS15h87wjpoXaY6ND4W/DeXter-UI---refresh?mode=dev

nguvictor:
I made the changes suggested by the both of you.
Wasn't sure about the text changing size.

EvgeniiaVak (approved):
yeah, changing size looked weird, thank you for not adding it 👍

PR #136, opened 2023-09-27
update to radix stokenet

EvgeniiaVak:


fliebenberg (approved):
Happy wiht the changes and youve confirmed that you could submit trades. SO merging.

PR #145, opened 2023-10-03
Charts enhancement

nguvictor:
Update Classes and tweaking padding, wrapped numbers in displayAmount.


src/app/components/PriceChart.tsx
```diff
@@ -4,30 +4,52 @@ import {
...
-        // TODO: timescale is not visible
+        height: 500,
+        //MODIFY THEME COLOR HERE
+        layout: {
+          background: { color: "#191B1D" }, //base-100
+          textColor: "#fff", //primary-content
+        },
+        //MODIFY THEME COLOR HERE
+        grid: {
+          vertLines: { color: "#292C34" }, //neutral
```
EvgeniiaVak, line 47: @nguvictor we don't use the neutral mostly, maybe base-200? also please review https://github.com/DeXter-on-Radix/website/pull/128 I'm updating colors there to reflect figma more.

Maybe use js import (import the js object from `tailwind.config.js` instead of hard-copying?

EvgeniiaVak (changes requested):

src/app/components/PriceChart.tsx
nguvictor: Base-200 wasn't visible even after adjusting to #191B1D found in #128 
image


------
primary-content was way too strong
image

------
secondary-content seems about right 
image

EvgeniiaVak (approved):

PR #147, opened 2023-10-07
Updating chart styles

### LINKED ISSUE #97: Price Chart styling

nguvictor:
Fixes #97

### DIFF
--- a/src/app/components/PriceChart.tsx
+++ b/src/app/components/PriceChart.tsx
@@ -2,7 +2,6 @@ import { createChart } from "lightweight-charts";
 import React, { useEffect, useRef } from "react";
 import {
   CANDLE_PERIODS,
-  OHLCVData,
   setCandlePeriod,
   handleCrosshairMove,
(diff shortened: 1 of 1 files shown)


EvgeniiaVak (changes requested):
I like volume and change added, looks like useful info. I would still add the date, so the user does not need to calculate it by looking to the x-axis.

The whole data row (ohlcv...) seems kind of mushed together, it's a bit hard to get what numbers belong to what label. Maybe making 2 rows and adding a bit more space between them?

What about colors to the numbers above the chart? They are for some reason both red and green in figma (maybe just to show what colors should be used) I think we had the logic for these numbers to also be red/green depending on the candle direction. Did you remove this on purpose?

nguvictor:
Made the tweaks suggested. 

Known issues: DisplayTime util, converts to local time, but on the volume graph it uses UTC. So this caused a discrepancy.

EvgeniiaVak (approved):

PR #159, opened 2023-10-16
save locale separators to store

### LINKED ISSUE #158: Read locale separators for displayAmount
Use user specific locale separators in displayAmount.

fliebenberg:
Fixes #158 

Have not had time to do extensive testing but for different locales, but it does read the locale separators and save them to the store to be used in displayAmount and elsewhere.

displayAmount will by default use the separators in the store, so no changes should be required in other code if separators was not specified explicitly.

### DIFF
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -8,6 +8,8 @@ import { Navbar } from "./components/NavBar";
 import { store } from "./redux/store";
 import { Provider } from "react-redux";
 import { usePathname } from "next/navigation";
+import { getLocaleSeparators } from "utils";
+import { uiSlice } from "redux/uiSlice";
 
 export default function RootLayout({
   children,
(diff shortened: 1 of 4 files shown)


abdulhaseeb13mar:
I only have the basic knowledge of testing libraries and unit testing. Haven't gone that deep into the testing. Try asking GPT.

EvgeniiaVak:
@abdulhaseeb13mar yes, I use gpt-4 and copilot - it's not omniscient unfortunately, it can't just analyze the repo structure and tell what the root problem is, it just helps finding solution to every next error, which is still tedious. I have a hunch there is just some incorrect config somewhere.

EvgeniiaVak (changes requested):
Fixing the tests proved to be quite time consuming (see #160 for more details, but basically it's because of import errors, we need to do it eventually of course some day), for the mvp, can we not involve redux into this? You can call `getLocaleSeparators` directly from inside the `displayAmount`. Having them in redux is helpful only if we create some kind of user settings (but we are not doing it for now). 

It should still work correctly without `useEffect` because this function (`toAmount`) will be called in the browser anyways as soon as we get some data from AlhpaDEX when the page loads (so even if default values would be created via ssg during build time in nodejs environment (which are mostly 0s anyways) they would still be overridden by the browser in live version).

So what I suggest is to just remove all the redux involvement in the separators.

abdulhaseeb13mar:
> So what I suggest is to just remove all the redux involvement in the separators.

i agree, as long as we are not storing user setting, there is not much need to store it in redux

EvgeniiaVak:
Created another PR: #164 without redux for separators, we can reopen this one if needed.

PR #164, opened 2023-10-22
No-state-locale-separators

EvgeniiaVak:
fixes tests and displayAmount usage
one test is still failing
because displayAmount logic seems to be incorrect


EvgeniiaVak:
This is an alternative approach to #159 - I just copied your commit, but removed redux to avoid testing import errors

One test still fails, but this time correctly - due to logic error in the displayAmount
```
    Expected: "1234"
    Received: "1.2K"
```

@fliebenberg I believe it would be easier for you to fix this error, you can push directly to this branch if you are ok with this approach.

fliebenberg:
I have added decimalSeparator and thousandsSeparator as 2 optional parameters to the displayAmount function. This adds nice functionality in case someone wants to be explicit about the separators, but more importantly it helps to make sure the tests can work. The problem in the test case you flagged was that although the test was written for cases where there is no thousandsSeparator, there was no way to explicitly force no thousandsSeparator. This is where the optional parameters in the function helps. All the tests work now again as expected.
I also just removed the uiSlice import from the store.js file.

fliebenberg (approved):
Everything seems to be working as expected now. displayAmount automatically uses locale specific separators unless the caller specifically provides separators.
All tests pass and app compiles without problem.

PR #171, opened 2023-10-27
update platform id

EvgeniiaVak:


fliebenberg (approved):
The benefit of having global constants! Nice and easy change. Happy to approve.

PR #195, opened 2023-11-09
sdk version updated to latest

fliebenberg:
This update is required as it contained a breaking change when connecting to the AlphaDEX backend. No further updates to dexter is required.


Radstakes (approved):
OK

PR #199, opened 2023-11-10
fixes separators logic disparities

### LINKED ISSUE #191: Decimals/thousands separator does not work correctly in OrderBook
bug report - https://t.me/dextertesters/64

I listed some at 50K XRD but orderbook shows as it was 50XRD (Same when i tryed 5k XRD).... Front end is using Periods(.) For both, decimmals and thousands (Price XRD collumm)

Image

EvgeniiaVak:
fixes #191 

- Removes disparities between separator logic for different components.
- Removes the separators params from displayAmount function - we don't want users to ever pass something non-standard here (something that is different than the rest of the page). They are not needed for tests, because just mock setup ensures consistent separators.
- Ensures there is always a thousand separator (because it simplifies code and on macs for some reason even when selecting a number format with spaces - the browser thinks there is no thousands separator, "").

### DIFF
--- a/__tests__/utils.test.ts
+++ b/__tests__/utils.test.ts
@@ -1,5 +1,7 @@
-import { displayAmount, displayNumber } from "../src/app/utils";
+import { displayAmount } from "../src/app/utils";
 
+// the separators are set to "." and " " for testing purposes
+// inside jest.setup.js
 describe("displayAmount", () => {
   it("sends error message if noDigits is less than 4", () => {
     let digits = 3;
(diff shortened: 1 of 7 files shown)


EvgeniiaVak:
@fliebenberg please take a look at these changes, especially to the FIXME comments - I think there is an error in the displayAmount logic that will happen for some users but not others

Also, maybe setting up Intl formatter instead of the custom implementation?

https://dev.to/gautemeekolsen/formatting-large-numbers-in-javascript-3am9#comment-266kk
> const formatter = Intl.NumberFormat('en', { notation: 'compact' });
> And then you can use
> const numberToDisplay = formatter.format(number)
> You'll get numbers like 1.7 T or 420 M.

nguvictor (approved):

PR #201, opened 2023-11-21
Reset state on pair switch

### LINKED ISSUE #193: Submit button is disabled in the Market Order Tab, if there is an "error" in the Limit Tab
bug report - https://t.me/dextertesters/111

Steps - Refresh the browser -> this results in ADEX1 - XRD pair displayed, switch to DEXTR - XRD -> Enter details in Market tab -> No Submit button activated -> Switch to Limit Tab -> Note warning displayed -> Enter something to remove the warning -> Switch back to Market Tab -> No order can be submitted

Image
Image

nguvictor:
Fixes #193 

Added some rest conditions when switching pairs to clear the states in the validationToken1 and validationToken2 as the states were carried when switching pairs.

### DIFF
--- a/src/app/redux/orderInputSlice.ts
+++ b/src/app/redux/orderInputSlice.ts
@@ -232,6 +232,14 @@ export const orderInputSlice = createSlice({
       state.tab = action.payload;
     },
     updateAdex(state, action: PayloadAction<adex.StaticState>) {
+      //This clears up any validation when switching pairs
+      state.validationToken1 = initialValidationResult;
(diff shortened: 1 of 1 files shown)


EvgeniiaVak (changes requested):
updateAdex fires whenever something in the adex state changes, it will happen even without user interacting with anything, so resetting the values here results in wiping out what user had previously inputed when the page is idle for a whie - we need to find another way of dealing with the errors, maybe separating the state for market and limit orders completely (but that's just a first thought, maybe there is a simpler solution)

PR #202, opened 2023-11-21
Decimals Mask

### LINKED ISSUE #167: Decimal places in amounts input validation are not always correct
From discussion in [discord](https://discord.com/channels/1125689933735145503/1125689934251032584/1166704408118181898):

chaotic 🦁 — 10/25/2023 3:46 PM
It now includes data about the divisibility and max decimals allowed for each token in a pair.
Currently (before the update) we have the information about price decimals for pairs adexState.currentPairInfo.priceMaxDecimals (which is set per pair), and adex.AMOUNT_MAX_DECIMALS (which is global). How does this update change this situation? Should we update something in our code? Maybe AMOUNT_MAX_DECIMALS is not global anymore, or something else?

Fred.xrd — 10/25/2023 3:51 PM
It is now a bit more complicated, as it is not guaranteed that token amounts can take 8 decimals. So we should replace the AMOUNT_MAX DECIMALS with the appropriate token max decimals as specified in the pairInfo fields token1MaxDecimals and token2MaxDecimals. Wherever the user needs to input an amount, we now need to use the appropriate limitation for decimals. Also when displaying amounts, we can round them to the appropriate decimals. Max decimals will always be 8 or smaller for all tokens. I think it should be fine for the MVP to just go with AMOUNT_MAX_DECIMALS as the transaction will just give an error message on the wallet if the user tries to put through a too small amount. We can then deal with this issue as a feature enhancement.

nguvictor:
Fixes #167

### DIFF
--- a/src/app/components/order_input/AmountInput.tsx
+++ b/src/app/components/order_input/AmountInput.tsx
@@ -59,7 +59,7 @@ export function AmountInput(props: TokenInputFiledProps) {
     if (debouncedValue) {
       console.log(debouncedValue);
       console.log(inputToken);
-      dispatch(orderInputSlice.actions.setAmountToken1(inputToken));
+      //dispatch(orderInputSlice.actions.setAmountToken1(inputToken));
     }
   }, [debouncedValue]);
 
--- a/src/app/components/order_input/AmountInput.tsx
+++ b/src/app/components/order_input/AmountInput.tsx
@@ -84,7 +84,7 @@ export function AmountInput(props: TokenInputFiledProps) {
           disabled={disabled || false}
           mask={Number}
           scale={targetToken.decimals}
-          radix={utils.getLocaleSeparators().decimalSeparator}
+          radix={"."}
           placeholder={placeholder}
           value={amount.toString()}
           className={
--- a/src/app/components/order_input/AmountInput.tsx
+++ b/src/app/components/order_input/AmountInput.tsx
@@ -22,7 +22,7 @@ interface TokenInputFiledProps extends TokenInput {
   payReceive: string;
   disabled?: boolean;
   onFocus?: () => void;
-  onAccept: (event: string) => void;
+  onAccept: (value: any) => void;
 }
 
 export function AmountInput(props: TokenInputFiledProps) {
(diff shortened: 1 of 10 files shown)


src/app/components/order_input/AmountInput.tsx
```diff
@@ -78,18 +79,20 @@ export function AmountInput(props: TokenInputFiledProps) {
           <span>{symbol}</span>
         </div>
 
-        <input
+        <IMaskInput
           disabled={disabled || false}
-          type="number"
+          mask={Number}
+          scale={targetToken.decimals}
+          radix="."
```
EvgeniiaVak, line 86: the dot here is to tell IMaskInput about the decimal separator? It might be different depending on user settings, so it probably won't work if the separator is `,`. We can fix this either by forcing the separator to always be `.` for decimals (and white space for thousands) - then change the `utils.getLocaleSeparators` function. Or use this function to get the separator.

EvgeniiaVak (changes requested):
And please remove or comment out the console.log statements (or change the eslint setting if you think we need them in deployed version). 
Overall all a big improvement in the usability, can't wait to merge it, currently the market tab is not very functional because of the errors.

src/app/components/order_input/AmountInput.tsx
nguvictor: Cleaned up the comments, but sorting out the different local separators might take a bit to debug.

EvgeniiaVak: you can change your OS settings, but you will also need to restart the browser after that

here is where it is on macos, on linux and windows it's also probably somewhere under language, time and numbers formatting
Screenshot 2023-11-27 at 00 15 25

EvgeniiaVak: if it causes a lot of problems you can either hardcode the `.` as a decimal separator and ` ` as thousands in `utils.getLocaleSeparators`, or just use whatever it returns now for `decimalSeparator` in your radix param (I vote for the second option, but in any case we need a single place of truth for the separator, let it be utils.ts, because why not)

nguvictor: It is causing a bunch of problems and really causing some serious scope creep.

I'm thinking it isn't a blocker for launch, but a good quality of life in the future.

Maybe we can resolve this in another issue in the future.

nguvictor:
Undid the attempt to change the decimal separator, as it caused a bunch of scope creep and bugs.

src/app/components/order_input/AmountInput.tsx
EvgeniiaVak: https://discord.com/channels/1125689933735145503/1125689934251032584/1179024092112883832

EvgeniiaVak (changes requested):
I can't enter `0` after `.` in market orders and I cannot enter `.` on limit orders at all.

EvgeniiaVak (approved):

PR #212, opened 2024-01-01
prototype env vars

### LINKED ISSUE #206: Environment variables for testnet and mainnet urls

nguvictor:
Fixes #206

### DIFF
--- a/next.config.js
+++ b/next.config.js
@@ -10,6 +10,10 @@ const nextConfig = {
   reactStrictMode: false,
 
   pageExtensions: ["js", "jsx", "mdx", "ts", "tsx"],
+  env: {
+    NETWORK: process.env.NETWORK,
+    DAPP_DEFINITION_ADDRESS: process.env.DAPP_DEFINITION_ADDRESS,
+  },
 };
 
 module.exports = withMDX(nextConfig);
--- a/next.config.js
+++ b/next.config.js
@@ -10,10 +10,6 @@ const nextConfig = {
   reactStrictMode: false,
 
   pageExtensions: ["js", "jsx", "mdx", "ts", "tsx"],
-  env: {
-    NETWORK: process.env.NETWORK,
-    DAPP_DEFINITION_ADDRESS: process.env.DAPP_DEFINITION_ADDRESS,
-  },
 };
 
 module.exports = withMDX(nextConfig);
--- a/src/app/subscriptions.ts
+++ b/src/app/subscriptions.ts
@@ -29,7 +29,7 @@ let subs: Subscription[] = [];
 
 export function initializeSubscriptions(store: AppStore) {
   let networkId;
-  switch (process.env.NETWORK) {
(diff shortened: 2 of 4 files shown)


EvgeniiaVak:
@nguvictor do you need help with fixing env vars for `.github/workflows/lint-and-test.yml` worflow for this PR?

EvgeniiaVak:
can we use only `NETWORK=stokenet` in the env vars (without `NETWORK_ID=2`), we can then specify in the code which ID to use? (will be easier to test by changing only one variable

next.config.js
```diff
@@ -10,6 +10,10 @@ const nextConfig = {
   reactStrictMode: false,
 
   pageExtensions: ["js", "jsx", "mdx", "ts", "tsx"],
+  env: {
+    NETWORK: process.env.NETWORK,
+    DAPP_DEFINITION_ADDRESS: process.env.DAPP_DEFINITION_ADDRESS,
+  },
```
EvgeniiaVak, line 16: Do we really need this doubling? next should be able to pick up the env vars by default

src/app/subscriptions.ts
```diff
@@ -48,7 +60,19 @@ export function initializeSubscriptions(store: AppStore) {
...
+      network = adex.ApiNetworkOptions.indexOf("mainnet");
+      break;
+    case "stokenet":
+      network = adex.ApiNetworkOptions.indexOf("stokenet");
+      break;
+    default:
+      network = adex.ApiNetworkOptions.indexOf("stokenet");
+  }
+
+  adex.init(adex.ApiNetworkOptions[network]);
```
EvgeniiaVak, line 75:
```suggestion
  adex.init(process.env.NETWORK || "stokenet");
```

EvgeniiaVak (changes requested):

src/app/subscriptions.ts
nguvictor: This was the first thing I tried, but ran into type issues.
The adex.init() is a bit strange it wants to have alphadexApiNetwork
image

image

next.config.js
nguvictor: Ran into an issue that they were undefined in browser. I followed a Stackoverflow said that you had to do this for browser/client side. Looks like the new docs suggest another strategy.
https://nextjs.org/docs/pages/building-your-application/configuring/environment-variables#bundling-environment-variables-for-the-browser

EvgeniiaVak: maybe using `NEXT_PUBLIC_` prefix for the env vars?

EvgeniiaVak (approved):
I tried locally with mainnet:

```env
NEXT_PUBLIC_NETWORK=mainnet # Options: mainnet or stokenet
NEXT_PUBLIC_DAPP_DEFINITION_ADDRESS=account_rdx168qrzyngejus9nazhp7rw9z3qn2r7uk3ny89m5lwvl299ayv87vpn5
```

works fine 👍

PR #215, opened 2024-01-19
tweaked z-index

nguvictor:
Fixes issue with Longer dropdowns, having interactive elements behind take priority causes some items to be unselectable


Radstakes (approved):
Works great on the deploy preview. Approved

EvgeniiaVak (approved):

PR #226, opened 2024-01-24
toString instead of toLocaleString

nguvictor:
This is workaround to get correct decimals to display, as the toLocaleString solution causes issues with number display
For example: 
image
image

The problem is how Number and toLocaleString() loses precision. Example in dev console
image

### DIFF
--- a/src/app/components/OrderBook.tsx
+++ b/src/app/components/OrderBook.tsx
@@ -61,19 +61,16 @@ function CurrentPriceRow() {
   );
 
   let spreadString = "";
+  let spreadValue = "";
 
   // checking for past trades here because adexState.currentPairInfo.lastPrice
   // is never null, and is = -1 if there were no trades
   let lastPrice = "";
   if (trades.length > 0) {
-    lastPrice = orderBook.lastPrice?.toString() || "";
-    //The full solution causes display issues as it will always display min digits
-    /*
-    lastPrice =
-      orderBook.lastPrice?.toLocaleString(undefined, {
-        minimumFractionDigits: token2MaxDecimals,
-      }) || "";
-      */
+    lastPrice = utils.displayNumber(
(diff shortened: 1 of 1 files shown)


Radstakes (approved):

fliebenberg:
Havent looked at the front-end code for some time, but why don't we use the displayAmount function here. We know the decimals for the price is fixed, so just display it the same way you display prices in the actual orderbook buys and sells.

EvgeniiaVak (changes requested):
window.innerWidth
1295

the clash still happens, I guess we need to split it in two rows when the containing div is narrow
Screenshot 2024-01-24 at 13 26 09

EvgeniiaVak:
I agree with using `utils.displayNumber` here, with something large for `noDigits` (so it wouldn't shorten the large prices with Ks and Ms, maybe 9?). I don't think we need to set up`fixedDecimals` param though, if we specify it, it will result in a lot of extra zeros for most pairs, which would not look pretty.

EvgeniiaVak:
> I think having the spread on a row below would be fine, or we could remove the spread value and just show the percentage?

even if we remove the spread value, if the price number is very long it might still overlap, moving spread to a row below (but only if they don't fit) would be better

src/app/components/OrderBook.tsx
```diff
@@ -86,26 +90,28 @@ function CurrentPriceRow() {
         2
       );
 
-      spreadString = `${spread} (${spreadPercent}%)`;
+      spreadString = `(${spreadPercent}%)`;
```
EvgeniiaVak, line 93: Maybe removing the parentheses?

```suggestion
      spreadString = `${spreadPercent}%`;
```

EvgeniiaVak (changes requested):

EvgeniiaVak (approved):

PR #229, opened 2024-01-26
Added default pair

nguvictor:
fixes #231

PR #230, opened 2024-01-26
Bumping Security Update

nguvictor:


EvgeniiaVak:
@nguvictor Is this a duplicate of #229 ? I see the same changes in the diff.

nguvictor:
My bad, might have included the change when I made the branch. It will technically update diff when default-pair is merged. I'll update this after that.

EvgeniiaVak (approved):

PR #242, opened 2024-02-01
fix buggy border on candle periods

### LINKED ISSUE #240: A not-good-looking border on chart period selector
Steps to reproduce (in chrome browser):
- select a new time period
- press `esc`

Image

dcts:
fixes #240 .

We could remove the border fully but it is considered bad practice as focus-visible styling is crucial for keyboard navigation, which is used by many people, including those with disabilities.

Before/After:
before-after


EvgeniiaVak (changes requested):
Is there a way to achieve the same in a more tailwind-like manner? Not adding a new custom class but rather using tailwind classes.

Also it would be good to have this border a little bit dimmed out, it's too white. Maybe making it yellow (like the underline, I think it's accent color, checkout daisyui colors we set here - https://github.com/dcts/website/blob/01dcaacac2177c51026597d811a3ba4a08b8537b/tailwind.config.js#L41-L66). Having both underline and border also looks a little bit strange 🤔 but we want to have the underline to indicate what is selected...

EvgeniiaVak:
@dcts I see that this PR is from your fork's main branch. While it doesn't create any issues for the original repo (https://github.com/DeXter-on-Radix/website) it might make it more difficult for you to keep your fork up to date. I advise you to create feature branches for PRs (like you did in #244 )

dcts:
@EvgeniiaVak adapted to tailwind like styles.

Concerning changing the style of the outline, we should be careful to keep a high contrast, as it otherwise would be contraproductive to accessibility:
accessibility-low-vision
(source: section [accessibility_concerns](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible#accessibility_concerns) from official documentation)

If we want to change the styling to brand colors I'd suggest to track that in a separate issue/PR as this would involve changing all `focus-visible` borders on the page to keep a consistent style overall.

Maybe an alternative solution is to find out why pressing `esc` triggers the `focus-visible` state to be activated in the first place and prevent that using JS, WDYT? 🤔 

And thanks a lot for the comment, I realized how inconvenient using the main branch is after trying to submit a second PR! 🤣 (I guess thats a reflex from using mercurial in the past, where you barely work with branches and always commit to main)

EvgeniiaVak:
@dcts the easiest way to fix the main branch for your fork is to:

1. create a separate branch with these changes and push it to the origin
2. revert your changes in main (in your fork) and force push it
3. create another PR from that new branch you created in step 1 to the upstream/main

Or we may try to continue with this PR... you might need to delete the repo / create another fork in the future to keep it updated... I'm not sure if git is capable of resolving such things (but maybe it is).

EvgeniiaVak (approved):

nguvictor (approved):

dcts:
@EvgeniiaVak awesome thanks for helping me out. I found a button "sync" in my forked repo and could automatically revert those commits, which also closed this PR. 

I re-uploaded this PR here: #248

PR #255, opened 2024-02-13
Set default candle period to 1D instead of 30m

dcts:
Given we currently have low liquidity, I suggest changing the default candle timeframe from 30m to 1D. Low liquidity chart might be off-putting and makes a bad first impression for new users IMO.

change-default-timeframe-to-1D

### DIFF
--- a/src/app/state/priceChartSlice.ts
+++ b/src/app/state/priceChartSlice.ts
@@ -14,6 +14,7 @@ export interface OHLCVData extends CandlestickData {
 }
 
 // Candle periods from alphadex:
+// -> reference: https://www.npmjs.com/package/alphadex-sdk-js?activeTab=readme#candleperiods-array
 // -> ['5m', '15m', '30m', '1h', '4h', '6h', '12h', '1D', '1W', '1M']
 export const CANDLE_PERIODS = adex.CandlePeriods;
 
@@ -28,7 +29,7 @@ export interface PriceChartState {
 }
 
 const initialState: PriceChartState = {
-  candlePeriod: adex.CandlePeriods[7],  // default to 1D candle period
+  candlePeriod: adex.CandlePeriods[7],  // defaults to 1D candle period
   ohlcv: [],
   legendCandlePrice: null,
   legendPercChange: null,
--- a/src/app/state/priceChartSlice.ts
+++ b/src/app/state/priceChartSlice.ts
@@ -13,6 +13,8 @@ export interface OHLCVData extends CandlestickData {
   value: number;
 }
 
+// Candle periods from alphadex:
+// -> ['5m', '15m', '30m', '1h', '4h', '6h', '12h', '1D', '1W', '1M']
 export const CANDLE_PERIODS = adex.CandlePeriods;
 
 export interface PriceChartState {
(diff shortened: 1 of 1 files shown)


fliebenberg:
It looks like all the tests are failing at the moment so have a look to fix that. Also, I am not sure exactly how the candle period is implemented in the dexter app, but you should ideally be changing the adex.currentCandlePeriod directly as that is used to determine what candle data to keep updated in the sdk. You should just be able to do it with adex,currentCandlePeriod = "1D", but again I am not sure how these varaibles are used in the code.

EvgeniiaVak (changes requested):
@dcts yeah, it looks like it's not working as expected, to reproduce:
1. go to the deploy preview, click nothing, hover on a candle and observe that it's still 30m despite 1D looks selected
Screenshot 2024-02-13 at 22 20 37
2. click on the 1D and the chart will update
Screenshot 2024-02-13 at 22 22 36

As Fred mentioned, it's because the actual data for the chart is fetched inside AlhpaDEX SDK (which has internal state), when a user selects a new period both redux state and AlphaDEX sdk states are updated here:

https://github.com/dcts/website/blob/fc6574278bba68761e2cc9032f21f6a5264a5048/src/app/state/priceChartSlice.ts#L113-L116

You might want to dispatch this action in a useEffect in the chart component.

dcts:
Thanks for the pointers @fliebenberg and @EvgeniiaVak , will look into it. (Lint error should be fixed!)

Concerning data I have a question: Is the data on localhost and in the deploy preview based on testnet transactions or mock data? Where is it coming from? (As I noticed it differs from data shown in the productive app on dexteronradix.com).

EvgeniiaVak:
@dcts oh yes, we need to add the env vars to readme / contributing.md at some point...

by default it's from testnet, but can be set to mainnet by using .env file in the root project dir:

either
```
NEXT_PUBLIC_NETWORK=stokenet # Options: mainnet or stokenet
NEXT_PUBLIC_DAPP_DEFINITION_ADDRESS=account_tdx_2_129kev9w27tsl7qjg0dlyze70kxnlzycs8v2c85kzec40gg8mt73f7y
NEXT_PUBLIC_DEFAULT_PAIR_ADDRESS=component_tdx_2_1crs8ud8rr680krgtlskauye7qnns5zdawdlspvcqceder6tysu884p
```

or 
```
NEXT_PUBLIC_NETWORK=mainnet # Options: mainnet or stokenet
NEXT_PUBLIC_DAPP_DEFINITION_ADDRESS=account_rdx168qrzyngejus9nazhp7rw9z3qn2r7uk3ny89m5lwvl299ayv87vpn5
NEXT_PUBLIC_DEFAULT_PAIR_ADDRESS=component_rdx1czgjmu4ed5hp0vn97ngsf6mevq9tl0v8rrh2yq0f4dnpndggk7j9pu
```

dcts:
@EvgeniiaVak thanks, I can successfully switch env's now. 

I followed the suggestion to dispatch the event in the PriceChart component. To streamline updates, I exported the initial state from `priceChartSlice.ts`, ensuring it now also updates the alphadex SDK. This means future changes to the initial state require updating only the `initialState` inside `priceChartSlice.ts`.

EvgeniiaVak (approved):

fliebenberg (approved):

PR #266, opened 2024-02-19
improve pair selector component

### LINKED ISSUE #223: UI improvements on dropdown padding
image

dcts:
Fixes #223.

Improved the pair selector by:
- replacing slash `/` and magnifying glass with chevron (as designed by @Kafkafrate)
- removed margin and made hover state take up all width (as mentioned in #223)
- sorted the coins by following algorithm: XRD/xUSDC, DEXTR/XRD, (rest sorted alphabetically)
- added coin icon to dropdown and emphasized the first token, making the second token less prominent (inspired by Binance)
- added hover state (70% opacity -> 100%opcity on hover)

# Before
Bildschirmfoto vom 2024-02-20 00-26-56

# After
Bildschirmfoto vom 2024-02-20 00-26-37

### DIFF
--- a/public/chevron-down.svg
+++ b/public/chevron-down.svg
@@ -1,0 +1,4 @@
+<svg width="26" height="26" viewBox="0 0 26 26" fill="none" xmlns="http://www.w3.org/2000/svg">
+<rect width="26" height="26" rx="2" fill="#1F2021"/>
+<path d="M12.6538 16.7077L7 11.0538L8.05383 10L12.6538 14.6L17.2538 10L18.3077 11.0538L12.6538 16.7077Z" fill="white"/>
+</svg>
--- a/src/app/components/PairSelector.tsx
+++ b/src/app/components/PairSelector.tsx
@@ -1,15 +1,34 @@
 import { useAppSelector, useAppDispatch } from "../hooks";
-import { selectPairAddress } from "../state/pairSelectorSlice";
+import { selectPairAddress, TokenInfo } from "../state/pairSelectorSlice";
 import { orderInputSlice } from "../state/orderInputSlice";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
-import { FaSearch } from "react-icons/fa";
+import Image from "next/image";
 
 interface PairInfo {
+  token1: TokenInfo;
   name: string;
   address: string;
 }
 function displayName(name?: string) {
-  return name?.replace("/", " - ").toUpperCase();
+  return name?.toUpperCase();
+}
+function getPairs(name?: string): string[] {
+  return name ? name.split("/") : [];
+}
+// Sort pairs as follows: 1) XRD/xUSDT (if present), 2) DEXTR/XRD (if present), then the rest alphabetically.
+function sortOptions(options: PairInfo[]): PairInfo[] {
+  const sortedOptions = options.sort((a: PairInfo, b: PairInfo) =>
+    a.name.localeCompare(b.name)
+  );
+  // Define first and second pair
+  const priorityPairs = ["XRD/xUSDC", "DEXTR/XRD"];
+  const priorityOptions: PairInfo[] = priorityPairs
+    .map((pair) => options.find((option) => option.name === pair))
+    .filter((val): val is PairInfo => val !== undefined);
+  const otherOptions = sortedOptions.filter(
+    (option) => !priorityPairs.includes(option.name)
+  );
+  return [...priorityOptions, ...otherOptions];
 }
 
 export function PairSelector() {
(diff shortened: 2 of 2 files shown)


EvgeniiaVak (changes requested):
@dcts maybe we can replace the `+` sign at the right with token icons? (both of them) I'm not sure if that was @Kafkafrate 's intention but maybe it was just a placeholder for the icons?

Without the backslash people won't know there is this shortcut for search (or did you remove the shortcut? I vote to get it back, I think it's a useful feature. Or replace with some other shortcut. Maybe we need to adjust its design, but I think it needs to be there.

dcts:
We discussed in Telegram to remove the `+` for now (updated the PR accordingly). We also discussed to replace it with a trading related metric, e.g. "liquidity of the trading pair". That would provide additional information for the user. But since the design and final decision is pending, I would track that in another issue.

Concerning the `/`: I personally think its rather confusing than helping. Until today I did not even realized that there is a possibility to search coins by pressing `/`. But the functionality is pretty cool tbh, I like it a lot (and did not remove it, so its still functioning). But I'd much rather have it as "easter egg" or hidden functionality, as its not vital for the overall user experience.

But thats just my opinion, happy to change if overall community sentiment is supportive of keeping it :)

nguvictor:
image

I think we agreed on using the Dexter pair as the first pair as it had the most volume on the chart.
image
VS
image

Kafkafrate:
I updated the dropdown in figma to include the search option for token. Should look better now. 

Screenshot 2024-02-20 at 16 09 38

dcts:
@EvgeniiaVak those examples you sent make a lot of sense for those sites, as they both have a main search bar at the top of the site. On dexter we don't have that, the input only appears once we click the pair selector, so its rather hidden.

I am not a fan of adding a tooltip on hover, as at that point, the cursor is already hovering over the pair, and with a single click the search input opens. Its faster than the shortcut. It's still usefull for educating the user, but the tooltip will always appear there and its a bit unneccessary.

A new suggestion might be to add a permanent hint in the main bar at the top next to the `CONNECT` button, for example like this (a designer would need to properly mock this up):
Bildschirmfoto 2024-02-20 um 19 31 30

This hint could also be positioned in the center, although might be too prominent.
Bildschirmfoto 2024-02-20 um 19 22 02

To summarize, we have the following options:
1. get rid of the arrow down, and bring back `/` and magnifying glass (or change style to match the more widely known "Type / to search" style). 
2. place a permanent hint in the main bar (left of connect button)
3. place a permanent hint in the main bar (in the center)
4. don't show any hint, treat the `/` as a hidden feature.

I personally support the last option. It reminds me a lot of github's "dot" shortcut (on any repository, you can press `.` to open an interactive vscode environment and actually run the project!). It's not advertised anywhere, but I heard it from a tiktok video, and use it every since. Truly a hidden feature thats badass, and knowing its hidden makes it even cooler. To me the shortcut search is exactly the same, an awesome hidden feature I don't want to miss, but no need to promote it prominently.

dcts:
Option 1 could be something like this (designer needed for proper mockup):
Bildschirmfoto 2024-02-20 um 19 59 07

dcts:
@nguvictor yes that makes sense! Adapted the PR so now `DEXTR/XRD` is the first pair.

dcts:
@EvgeniiaVak added a permanent hint as proposed in Option 1.

dcts:
@EvgeniiaVak good call! Fixed it, now the `/` is also working inside the search input. 

What do you think about the permanent shortcut hint that is currently placed on top right of the screen? Happy to change if you have a strong preference towards another option. If we have no clear solution we could also consult with the community via poll on telegram.
permanent-hint

EvgeniiaVak:
@dcts my strong opinion, but very loosely held 😄 I'd prefer:
1. the way it is on mainnet now - the shortcut shown directly in the input field
2. or hidden if the first option for some reason is bad

if we keep the hint where it is, it probably should be less rounded?

about text colors - I noticed even the token1 color is darker now, is it from a design file? (looks worse in my opinion) 
Screenshot 2024-02-22 at 15 13 16

background for the down arrow looks a bit off

we probably need `cursor-pointer` on hovering this element because it's clickable

Do we have a related design file somewhere? Can we pin it in description here?

dcts:
@EvgeniiaVak ok sounds good. I'll ask in Telegram for a designer to take a look into this and iterate over current design (to make it more clear). For that goal, would you mind if we change the shortcut to for example Ctrl+K (CMD+K)? I think having 2 keys shown will make it very clear (see my previous suggestion, but ultimately up to the designer):
> Option 1 could be something like this (designer needed for proper mockup): Bildschirmfoto 2024-02-20 um 19 59 07

The design file I used is this one (but @Kafkafrate adapted the designs recently, so you might not find the old designs, but I am actively working on the update):
https://www.figma.com/file/qushS15h87wjpoXaY6ND4W/DeXter%20UI%20-%20refresh?type=design&fuid=708572579352018658

dcts:
Adapted to @Kafkafrate's new design:
- 2 icons showing both coins
- added price on the right side. Did not add 24h price change as all coins besides DEXTER have invalid values (-1). I suspect this means no price change in last 24h for most coins. I suggest leaving only the price there for now and implement future improvements in a seperate CL, as we are not aligned on it yet.

@EvgeniiaVak I reverted the chevron-down to what we currently have.

EvgeniiaVak (approved):

nguvictor (approved):

PR #270, opened 2024-02-23
fix empty chart canvas

### LINKED ISSUE #269: Chart real estate not fully used
The chart looks a bit empty, as there are currently few candles. To properly make use of the screens real estate, I suggest to automatically zoom in.

Current Behavior:
Bildschirmfoto vom 2024-02-23 15-17-30

Expected Behavior:
Bildschirmfoto vom 2024-02-23 15-17-21

dcts:
Fixes #269.

Just a proposal, I think it looks better when the candles are zoomed and real estate of the chart is fully used.

### DIFF
--- a/src/app/components/PriceChart.tsx
+++ b/src/app/components/PriceChart.tsx
@@ -169,6 +169,11 @@ function PriceChartCanvas(props: PriceChartProps) {
 
       window.addEventListener("resize", handleResize);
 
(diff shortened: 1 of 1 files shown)


nguvictor:
I think that this is a great improvement, it's only possible now that the project has enough data to support the minimum amount of days to fill out the chart

fliebenberg (approved):
Nice change. We will need to revisit this in future when there is too much data, but that would be a good problem to have.
Happy to approve.

EvgeniiaVak (approved):

PR #277, opened 2024-03-02
Fix missing DEXTR icon

dcts:
Currently the DEXTR icon seems to be not available on IPFS. I experienced this before. I suggest hardcoding it and using coingecko's hosted png. It should have a more reliable uptime. 

Bildschirmfoto vom 2024-03-02 23-39-38


dcts:
After some digging Faraz and I found out:
- icon shows normally for most people (works also on my android phone and on my mac laptop)
- seems to happen only on linux

Can any linux user confirm whether this happens to you too?

dcts:
This issue only occurs when my VPN is turned on.

fliebenberg:
I dont ever see the DEXTER token when I look at the previews for merge requests. Not sure if that helps to track down the problem. It might be better to just use the coingecko png if we can. Is only th elogo hosted on IPFS. I thought the whole site is hosted there?

Radstakes:
No it's just the logo that is hosted on IPFS currently, but the logo displays for me on both deploy previews and main releases.

EvgeniiaVak:
I also always see the logo: on stokenet, on mainnet, locally (no vpn, macos, chrome)

EvgeniiaVak (changes requested):
We keep these iconUrls in redux stores. Maybe assigning the correct url where the assignment from adex state to redux takes place?

https://github.com/dcts/website/blob/f3774a9b817da225b18adfd040d33f7f1aaeccfb/src/app/state/orderInputSlice.ts#L273-L291

https://github.com/dcts/website/blob/f3774a9b817da225b18adfd040d33f7f1aaeccfb/src/app/state/pairSelectorSlice.ts#L97-L105

For now it's still 4 places, but this way if we reuse this state we don't need to remember putting this if/else in components 🤔 although maybe we won't be reusing them... Anyways, it sounds more appropriate to me to keep this logic together with where we fetch the data.

If for some reason still keeping the if/else inside the components, I think there is one missing here:
https://github.com/dcts/website/blob/f3774a9b817da225b18adfd040d33f7f1aaeccfb/src/app/components/PairSelector.tsx#L223-L227

dcts:
Good call, I agree, changing the urls inside components feels wrong. Will adapt. But to fix the state for the pair_selector component, I need to iterate over all options, but I am not sure where those options are coming from, for example:

On [this line](https://github.com/dcts/website/blob/f3774a9b817da225b18adfd040d33f7f1aaeccfb/src/app/components/PairSelector.tsx#L39
) we get 
```
const pairSelector = useAppSelector((state) => state.pairSelector);
```

And `pairSelector.pairsList` holds all options. But when I try to iterate over them and change all iconUrl for DEXTR:
```
for (let i = 0; i < pairSelector.pairsList.length; i++) {
    if (pairSelector.pairsList[i].token1.symbol === "DEXTR") {
      pairSelector.pairsList[i].token1.iconUrl =
        "https://assets.coingecko.com/coins/images/34946/standard/DEXTRLogo.jpg";
    }
    if (pairSelector.pairsList[i].token2.symbol === "DEXTR") {
      pairSelector.pairsList[i].token2.iconUrl =
        "https://assets.coingecko.com/coins/images/34946/standard/DEXTRLogo.jpg";
    }
  }
```

I get this error:
```
TypeError: Cannot assign to read only property 'iconUrl' of object '#<Object>'
```

What would be the correct place to adapt this state given its readonly? @EvgeniiaVak

EvgeniiaVak:
@dcts JS passes arguments by sharing (a function can mutate the original object) so we can create a funtion (maybe in utils) that takes the whole token object and put the if/else there, later we still need to assign the result in the state (so redux would update it properly).

---

> its still up for debate whether we use coingecko URLs or not. I am still experiencing it on VPN but looks like I am the only one affected.

Converting this PR to draft for now, so people wouldn't accidentally merge it.

dcts:
Adapted changes to use helper function.

EvgeniiaVak (approved):

EvgeniiaVak:
@fliebenberg does this PR fix things for you on testnet?

fliebenberg:
> @fliebenberg does this PR fix things for you on testnet?

Yep. When I tested with the deployment you sent on Telegram it worked.

EvgeniiaVak:
@dcts  I think it's good to go in, looks like it fixes some issues. It still doesn't solve them when VPN is on (getting it from your previous comment, I'm not using VPN), is that correct?

dcts:
@EvgeniiaVak Got it! 

My previous comment might have been misleading, for me this PR solved all my issues with missing icons, especially when VPN is on.

EvgeniiaVak:
We need a second approval on this PR, @fliebenberg  could you grant yours?

fliebenberg (approved):

PR #279, opened 2024-03-03
fix price in trade history

### LINKED ISSUE #234: Order price in Open Orders table showing wrong token name
In the Open Orders (and also Order History) table the Order Price column shows the wrong token symbol.

image 
In the example above the price should be in XRD, not DEXTR.

dcts:
Fixes #234. 

Previously the token price was determined by the unclaimed token, yet in some cases this lead to an error. The price should always be expressed in the second token pair, independent of side or specified token.

# All 4 usecases
Bildschirmfoto vom 2024-03-03 22-17-56

# Before
beforree

# After
afterrr


fliebenberg (approved):
Happy to approve.

src/app/utils.ts
```diff
@@ -292,3 +292,12 @@ export function numberOrEmptyInput(event: string) {
   }
   return amount;
 }
+
+// Given an order, determine token symbol in which the price is expressed
+// Note: Price is always expressed in terms of the second currency in the trading pair.
+export function getPriceSymbol(order: any): string {
```
EvgeniiaVak, line 298: Please add proper type instead of `any` here.

```TS
import type { OrderReceipt } from "alphadex-sdk-js";
...
...order: OrderReceipt...
```

EvgeniiaVak (changes requested):
@fliebenberg do you think it would be useful to add price symbol field to AlphaDEX SDK (OrderReceipt interface)?

fliebenberg:
> @fliebenberg do you think it would be useful to add price symbol field to AlphaDEX SDK (OrderReceipt interface)?

I tried to keep the order receipt info as small as possible as it is all stored on-ledger. The price symbol will always be token2, so it should be easy to get from the pair info associated with the order.

src/app/utils.ts
dcts: Done.

EvgeniiaVak (approved):

PR #286, opened 2024-03-14
Fix trade history bug

dcts:
Looking at @Kafkafrate screenshot (bugreport from telegram: https://t.me/dexter_discussion/12674):

image
image

First of all, if the order was filled 100% then "Filled qty" and "Order qty" should have the same quantity and token.

Given the price is 2 XRD, this means that 5XRD were paid to buy 2.5 DEXTR. This leads me to believe, that "Filed Qty" is refering to amount in DEXTR, while "Order Qty" is referring to amount in XRD. 

To ensure both coins are always the same, this PR calculates the "Filled Qty" from scratch, which should solve this inconsistency.

(side note: I have a suspicion that [`amountFilled`](https://socket.dev/npm/package/alphadex-sdk-js/files/0.12.9/lib/models/order-receipt.d.ts#L35) is not always specified in the same token as [`amount`](https://socket.dev/npm/package/alphadex-sdk-js/files/0.12.9/lib/models/order-receipt.d.ts#L28), can you confirm or deny @fliebenberg?)

### DIFF
--- a/src/app/components/DisplayTable.tsx
+++ b/src/app/components/DisplayTable.tsx
@@ -25,8 +25,8 @@ const headers = {
     "Order Type",
     "Direction",
     "Time Ordered",
-    "Order Price",
     "Amount",
+    "Order Price",
     "Filled Qty",
     "Completed %",
     "Action",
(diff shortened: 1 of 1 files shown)


nguvictor (approved):

fliebenberg:
amount and filled quantity should always refer to the specified token.
I am aware of a bug in the filled amount for some orders not updating on the api database sometimes. I have not had time to look for the reason for this, but it is on my list. Your fix of using percent completed should hide it from the user for now. The amount filled is correct on the actual receipt on the ledger, it is just the API that does not update its database correctly when an open order gets filled.

fliebenberg (approved):
Happy.

PR #299, opened 2024-03-20
add adapting page title

dcts:
Currently the tab only shows "DeXter". This PR adds adapting pagetitles to also show the currently selected pair in the tab, which adds additional context and value to the user. 

Bildschirmfoto vom 2024-03-20 11-41-22


fliebenberg:
I love this change.  I would suggest you use the name field from the pair rather than recreating the name manually using the tokens. It is possible that the name of the pair is changed to not follow the standard format of token1/token2 in future and this will be reflected in the name field of the pair.

nguvictor (approved):

src/app/utils.ts
```diff
@@ -346,3 +346,12 @@ export function detectBrowserLanguage(defaultLanguage: string = "en"): string {
...
 }
+
+// If available, adds currently selected pair to the page title.
+export function updatePageTitle(pairName: string) {
+  if (pairName === "") {
+    document.title = "DeXter";
+  } else {
+    document.title = `DeXter • ${pairName}`;
+  }
+}
```
EvgeniiaVak, line 357: Will this logic be used only once in the trading page? If yes, it's better to keep it there, not to grow the utils indefinitely. If it's for all pages then maybe editing the head->title tag in `src/app/layout.tsx`? 

If it will be useful for several pages but not for all, the name of the function should be something different not `updatePageTitle`,  maybe `addPairToTitle`. Or if improving reusability a bit more - create something like a custom hook instead of a util function, so that that pairSelector would also go there and if a page needs the pair name in the title it would just call `useTitlePageWithPair()` or something like this.

EvgeniiaVak (changes requested):

src/app/utils.ts
dcts: Good questions. I agree the util seems unneccessary, removed it and kept the logic directly inside the trade component.

Concerning using the pair in the title throughout the webapp: I don't think that is planned. The pair should only be shown in the "Trade" page. While I do like the idea of creating a custom hook, I think a pragmatic solve here is sufficient :)

Let me know what you think!

EvgeniiaVak (approved):

src/app/utils.ts
EvgeniiaVak: just in /trade page is perfect 👍

PR #313, opened 2024-03-25
Create stokenet-deploy-main.yml

EvgeniiaVak:
This is NOT for PR previews, it is deploying main branch with stokenet env to github pages.

Default workflow for next.js recommended by GitHub itself. We have default testnet env vars set up directly in the code as a fallback if no env vars provided, so should without specifying them.


Radstakes (approved):
Ok

nguvictor (approved):

PR #320, opened 2024-03-30
migrate from jest to vitest

### LINKED ISSUE #317: Failed to setup tests for OrderInputSlice
When I try to setup tests for OrderInputSlice (see branch: [orderInputSlice-test](https://github.com/DeXter-on-Radix/website/tree/orderInputSlice-test) -> see [code](https://github.com/DeXter-on-Radix/website/compare/main...orderInputSlice-test)), I get the an error (see below):

After doing some research it seems like nextJS is enabling ECMAScript Modules by default, setting up babelrc file did not help.

## Error
```bash
 PASS  __tests__/i18nSlice.test.ts
 PASS  __tests__/utils.test.ts
 PASS  __tests__/orderBookSlice.test.ts
 FAIL  __tests__/orderInputSlice.test.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /home/dcts/code/dexteronradix/website/node_modules/lit/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import"@lit/reactive-element";import"lit-html";export*from"lit-element/lit-element.js";export*from"lit-html/is-server.js";
                                                                                      ^^^^^^

    SyntaxError: Cannot use import statement outside a module

      20 | let rdtInstance: null | RDT = null;
      21 | export function getRdt() {
    > 22 |   return rdtInstance;
         |                           ^
      23 | }
      24 | function setRdt(rdt: RDT) {
      25 |   rdtInstance = rdt;

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1505:14)
      at node_modules/@radixdlt/radix-dapp-toolkit/dist/radix-dapp-toolkit.umd.js:1:172
      at Object.<anonymous> (node_modules/@radixdlt/radix-dapp-toolkit/dist/radix-dapp-toolkit.umd.js:1:662)
      at Object.<anonymous> (src/app/subscriptions.ts:22:27)
      at Object.<anonymous> (src/app/state/orderInputSlice.ts:69:24)
      at Object.<anonymous> (src/app/state/store.ts:12:26)
      at Object.<anonymous> (__tests__/orderInputSlice.test.ts:5:16)

Test Suites: 1 failed, 3 passed, 4 total
Tests:       15 passed, 15 total
Snapshots:   0 total
Time:        1.672 s
Ran all test suites matching /__tests__/i.

```

EvgeniiaVak:
fixes #317 

- moved the test files from __test__ folder to src as it kind of make sense to keep them very close, and hopefully people will write more unit tests together with new features 😃 
- the test still fails, but not because of env set up - there is legitimate "" does not equal 0
- will probably need to quit/launch vscode a couple of times for the testing tabl to work correctly
- replaced a variable MAX_ROW in test names because vscode testing tab could not handle that for vitest (although npm run test worked all right)


src/app/state/orderBookSlice.test.ts
```diff
@@ -40,19 +40,19 @@ const MOCK_BUYS = [
 ];
 
 describe("toOrderBookRowProps", () => {
-  it(`returns ${MAX_ROWS} rows if the input is empty`, () => {
+  it(`returns 11 rows if the input is empty`, () => {
```
dcts, line 43: Orderbook rows amount will change in the recent redesign. I would prefer not hardcoding the number. 
Suggestion: "returns correct amount of rows if input is empty".

src/app/state/orderBookSlice.test.ts
```diff
@@ -40,19 +40,19 @@ const MOCK_BUYS = [
...
     const input: OrderbookLine[] = [];
 
     expect(toOrderBookRowProps(input, "sell", 0).length).toBe(MAX_ROWS);
   });
 
-  it(`returns ${MAX_ROWS} rows if the input is smaller`, () => {
+  it(`returns 11 rows if the input is smaller`, () => {
```
dcts, line 49: same here.
"returns correct amount of rows if the input is smaller"

src/app/state/orderBookSlice.test.ts
```diff
@@ -40,19 +40,19 @@ const MOCK_BUYS = [
...
     const input: OrderbookLine[] = MOCK_SELLS.slice(0, 5);
 
     expect(toOrderBookRowProps(input, "sell", 0).length).toBe(MAX_ROWS);
   });
 
-  it(`"returns ${MAX_ROWS} rows if the input is larger"`, () => {
+  it(`"returns 11 rows if the input is larger"`, () => {
```
dcts, line 55: same here.
"returns correct amount of rows if the input is larger"

dcts (changes requested):
Just minor change, otherwise looks great! Thanks for migrating! 🙏💯

PR #322, opened 2024-04-01
Calculator service

dcts:
Given dexter is a financial application where calculations are crucial, we should not rely upon javascripts floating point arithmetic, as it has some downsides:

image

This PR introduces a Calculator service that should be used for all calculations going forward.


EvgeniiaVak (commented):
This is cool! We probably will need a couple of PRs after that to fix any math we have so far to use it. 

One thing code-style-wise is bugging me though, just esthetics, if all the methods a static, why having a class at all, maybe just the set of functions?

nguvictor (approved):

dcts:
> This is cool! We probably will need a couple of PRs after that to fix any math we have so far to use it.
> 
> One thing code-style-wise is bugging me though, just esthetics, if all the methods a static, why having a class at all, maybe just the set of functions?

Agree this could be implemented using a more functional esthetic, thanks for the question! It would have bugged me probably after some time too! :)

fliebenberg (approved):

dcts:
@EvgeniiaVak merging with removed class structrue, if you have any more comments feel free to create another PR or ping me on discord :)

PR #327, opened 2024-04-03
Add toaster notifications (for submit order + cancel order)

dcts:
Implemented toast notifications for the following user actions:
- submit order (market or limit)
- cancel order

The toast is waiting in a loading state, until the user either accepts or rejects the transaction in the wallet app.

### DIFF
--- a/src/app/components/DisplayTable.tsx
+++ b/src/app/components/DisplayTable.tsx
@@ -69,32 +69,26 @@ function ActionButton({
   if (order.status === "PENDING") {
     return (
       <button
-        onClick={(e) => {
+        onClick={async (e) => {
           e.stopPropagation();
+          // Execute DexterToast.promise with the provided success and failure messages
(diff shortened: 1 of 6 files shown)


dcts:
Added conditional display of the toasts based on operating system, to ensure that (1) notification from browser and DeXter are not overlapping and (2) both notifications are in close proximity. 

Results:
- on **linux**: browser notifications are top center, hence show toasts in top-center (they do not overlap)
- on **windows**: browser notifications are bottom right, hence show toasts in bottom-center
- on **mac**: no browser notifications, but theoretically they should be top right, hence show toasts in top-center

Bildschirmfoto 2024-04-04 um 00 56 19
Bildschirmfoto 2024-04-04 um 00 56 26
Bildschirmfoto 2024-04-04 um 00 56 33

nguvictor (approved):

PR #330, opened 2024-04-05
fix pair component missing unique key bug

dcts:
Nothing major, but having unique keys specified for each elements in a loop improves performance. 

image

I'm surprised that next lets us export the app with this error, I believe it would be better to block deployment in such a case, WDYT @EvgeniiaVak?

### DIFF
--- a/src/app/components/PairSelector.tsx
+++ b/src/app/components/PairSelector.tsx
@@ -190,7 +190,7 @@ export function PairSelector() {
           const [pair1, pair2] = getPairs(option["name"]);
 
           return (
-            <React.Fragment key={`fragment-${index}`}>
+            <React.Fragment key={`pair-${index}`}>
               {index === 0 && (
                 <div className="flex justify-between text-sm opacity-40 ml-3 mr-3">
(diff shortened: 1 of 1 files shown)


src/app/components/PairSelector.tsx
```diff
@@ -189,7 +190,7 @@ export function PairSelector() {
           const [pair1, pair2] = getPairs(option["name"]);
 
           return (
-            <>
+            <React.Fragment key={`fragment-${index}`}>
```
EvgeniiaVak, line 193:
```suggestion
            <React.Fragment key={`pair-${index}`}>
```
If we use fragment for some other indexed lists on the same page, we will have the same error, pair- prefix would be less generic.

EvgeniiaVak (changes requested):

EvgeniiaVak:
@dcts yes, blocking deployment on such warnings would be nice

EvgeniiaVak (approved):

nguvictor (approved):

fliebenberg:
Agree. Most languages these days warn you about this. So agree to block deployment if not present.

PR #335, opened 2024-04-08
handle isConnected state

dcts:
This PR adds the users login status (logged in vs not logged in) to the radix slice, which is a useful feature currently needed for the rewards page.

Its a very simple change so I am a wondering why it was postponed and marked as TODO in the first place @EvgeniiaVak , am I missing some issues here?

### DIFF
--- a/src/app/state/radixSlice.ts
+++ b/src/app/state/radixSlice.ts
@@ -16,7 +16,6 @@ const initialState: RadixState = {
     personaData: [],
     proofs: [],
   },
-  // TODO: handle connection status
   isConnected: false,
 };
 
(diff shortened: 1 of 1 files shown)


EvgeniiaVak (approved):
No, you are not missing anything I think, no logic depended on that state at the time that's why it wasn't implemented.

One question, not a blocker at all, why having it as a separate PR if the first need for it is in rewards component, maybe just add it in rewards component PR?

dcts:
> No, you are not missing anything I think, no logic depended on that state at the time that's why it wasn't implemented.

Got it, thanks for clarifying :)

> One question, not a blocker at all, why having it as a separate PR if the first need for it is in rewards component, maybe just add it in rewards component PR?

I prefer this flow for several reasons:
- Rewards PR might not be merged very soon while this feature is ready
- reduces complexity from rewards PR 
- rewards PR csn simply merge from main to have this feature

fliebenberg (approved):

Issue #314, opened 2024-03-25
Add "Report Bug" feature


dcts:
We have a good system where bugs can get reported on TG or Github and usually get picked up by devs.

But I think we miss out on feedback from users that are not on TG nor on Github. I suggest to add "Report Bug" action in the footer (that sends an email to our official account). 

Alternatively we could integrate a tool like Hotjar (https://www.hotjar.com/product/feedback/), although it might be overkill.

dcts:
fixed in #343

Issue #304, opened 2024-03-20
Setup Flow to report translation issues


dcts:
I set up the following document to report and tackle translation issues:

https://docs.google.com/document/d/11I9c3XjKPmXgc9V6puznD_mSXA5QCqNKAG1QCcClkRw/edit?usp=sharing

My proposal would be to add a link in the footer "report translation issue", for easy access, or alternatively, create dexteronradix/report path that redirects to the doc and post this in the TG. But I believe the first option is better as the user does not need any prior knowledge on where to report translation issues.

Related issues: #303 #301

dcts:
Done.

dcts:
fixed in #343

Issue #303, opened 2024-03-20
Add more content in footer


dcts:
Proposal to add more content in the footer (DESIGN pending):
(this suggestion is inspired by pancakeswap, see discussion https://github.com/DeXter-on-Radix/website/issues/265#issuecomment-1959812154)

Some examples:

ECOSYSTEM
Trade -> links to https://dexter-on-radix.gitbook.io/dexter/using-dexter/start-trading
Provide Liquidity -> links to https://dexter-on-radix.gitbook.io/dexter/overview/how-are-contributors-rewarded/liquidity-incentives
Tokenomics -> links to https://dexter-on-radix.gitbook.io/dexter/overview/how-are-contributors-rewarded/tokenomics

DEVELOPERS:
contribute -> links to https://dexter-on-radix.gitbook.io/dexter/overview/how-do-i-contribute
github -> link to https://github.com/DeXter-on-Radix

SUPPORT
send us a message -> link to https://t.me/dexter_discussion
report translation issue -> link to https://docs.google.com/document/d/11I9c3XjKPmXgc9V6puznD_mSXA5QCqNKAG1QCcClkRw/edit?usp=sharing
documentation -> link to https://dexter-on-radix.gitbook.io/dexter

BLOG
founding article -> link to the founding medium article
tokenomics -> link to tokenomics blogpost by Arc

Adding a bit more content in the footer would help, because:
- it adds "shortcuts" and content to explore
- improves UX and keeps users engaged
- helps with SEO, as we provide better context for crawlers and index important keywords

dcts:
# Some inspiration

**Binance**
![image](https://github.com/DeXter-on-Radix/website/assets/44790691/c2d21c28-9011-4c25-8166-e4199e72e405)

**Pancakeswap**
![image](https://github.com/DeXter-on-Radix/website/assets/44790691/42fd8f83-5637-4274-8d20-4eaf2bed5559)

dcts:
fixed in #343

PR #343, opened 2024-04-13
Footer Redesign

### LINKED ISSUE #265: the "trade" link at the bottom of the website
It looks like the "trade" link at the bottom of the website and the title of the social media channels under this link. For this reason, the trade link should be removed from here
Social Media Channels   Trade link

dcts:
Added a new footer:
- fixes #265, #303, #304, #314
- added translations
- multi column layout that wraps automatically to ensure responsiveness
- bigger social media icons
- more content to explore
- logo now green instead of white
- removed "runs on radix" button, as not consistent with dexter branding

**Before**
image

**After**
image


src/app/styles/globals.css
```diff
@@ -121,3 +121,9 @@ input[type="number"]::-ms-clear {
 .custom-auto-row-grid {
   grid-auto-rows: 0fr auto 0fr;
 }
+
+h1,
+h2,
+h3 {
+  margin: 0 !important;
```
nguvictor, line 128: We want to avoid using custom classes would it be possible to do something like this?
!m-0

https://tailwindcss.com/docs/configuration#important-modifier

nguvictor (changes requested):

src/app/styles/globals.css
dcts: This is not really a custom class, but more a global style applied to all headers. I decided to include it in global styles as I noticed I usually remove margins just to add customized margins back for all headings I use. We also have global styles for other usecases. 

But if you prefere I can either (1) remove and use !m-0 as you suggested or (2) add it to an already specified global class (see attached image):

IMG_20240414_125926

@nguvictor

dcts: added the heading to an already existing global style as described in my previous comment (see screenshot). LMK if you have concerns still, happy to change and use `!m-0` if you have any objections on why not to use global class for this usecase :)

nguvictor:
Smaller width devices, look a bit strange, but could be polished as a different PR to polish for mobile.
image
image
image

nguvictor (approved):

fliebenberg (approved):

PR #346, opened 2024-04-14
Navbar redesign

### LINKED ISSUE #310: Title component unexpectedly appears on small screensizes
When screenwidth is less than 320px the title "DEXTER" suddently appears again and overlaps with other content. 

Bildschirmaufnahme2024-03-21um10 00 22-ezgif com-video-to-gif-converter

Ideally the title should not be shown for screenwidths less than 320px.

dcts:
Update design for navbar:
- add same logo as in footer according to design file
- reduce fontsize
- add responsiveness: mobile screens have hamburger menu and a simple fullscreen menu
- mobile version does not include a radix connect button, as currently radix doesnt support mobile clients logging in

**Before**
Bildschirmfoto vom 2024-04-15 20-49-19

**After**
Bildschirmfoto vom 2024-04-15 20-40-53

**Mobile Menu**
Bildschirmfoto vom 2024-04-15 01-53-55

This PR fixes #310.


fliebenberg (approved):

fliebenberg (approved):

fliebenberg (dismissed):

EvgeniiaVak (changes requested):
`Connect` button is gone even if I'm on laptop but very narrow screen, we probably only want to remove it on mobile. And we need to say why it's not there, something like "Connection is not yet supported for mobile devices".

Why is there the word "Trade" in the menu? is it the page title? Maybe it should be something like "Settings"?

dcts:
@EvgeniiaVak I'm currently using the smallest breakpoint from [tailwindcss](https://tailwindcss.com/docs/responsive-design) (640px for Mobile):
image

So currently the button is only gone for screensizes < 640px. Should I create a custom breakpoint at 500px for example?

Concerning "Trade", yes that is the page navigation. Since we only have 1 page it might be confusing, but since we'll launch rewards soon this will automatically resolve itself.

EvgeniiaVak:
> Concerning "Trade", yes that is the page navigation. Since we only have 1 page it might be confusing, but since we'll launch rewards soon this will automatically resolve itself.

I think having a page name in the settings is a bit strange anyways, even when we have multiple pages. The user selects language once, not separately for every page, right? Maybe not "Settings", but "DeXter" instead of a page name?

EvgeniiaVak:
> So currently the button is only gone for screensizes < 640px. Should I create a custom breakpoint at 500px for example?

No, I meant we need to find a way to disable the connect button only on mobile devices, but not on narrow desktop windows. Maybe using the new isMobile from utils? (Btw right now it is used only to decide a width of something? Why not tailwind prefixes there?)

fliebenberg (approved):

EvgeniiaVak (approved):

PR #347, opened 2024-04-15
Add promo banner (Validator Node Staking)

dcts:
Add capability of displaying promo banners on the Trade page of dexter. 

### To add a banner:
1. define `imageUrl` and `imageUrlMobile` inside `src/app/layout.tsx` 
2. optionally define a `redirectUrl` for when the banner is clicked
3. the banner will automatically show if both images are provided

## Features:
- the banner display logic is different depending on screensize:
  - `< 500px`: show small image (500x128), and scale down linearly
  - `> 500px`: show large image (1640x128) without scaling, but center content
  - `> 1640px`: show large image (1640x128) without scaling, and add gradient colors on left and right side to prevent banner cutoff
- added rudimentary banner for validator node staking

## Remark on handling resize
As the implementation requires the correct image to be present (e.g. the implementation breaks if a large image is loaded for screensizes <500px), this cannot be done with CSS and srcset attribute alone. Hence I chose to handle the resize and image loading with javascript.


Radstakes (approved):
OK

fliebenberg (approved):

PR #348, opened 2024-04-15
Promo banner improvements

dcts:
**Improvements**
- gradient is handeled by the website to simplify flow
- banner content is provided from banner designer as SVG with transparent background
- new aspect ratios: 
  - 600x80 for desktops
  - 600x200 for mobile (for wider display)
- added documentation on how to add/change banners
- added new redirectUrl for validator staking promo banner


EvgeniiaVak (changes requested):
A thing that I found weird about the current banner is that "Learn more about how to stake" looks like a button that would lead to some documentation (I was expecting gitbook link there) but instead it opens the same page as when you click anywhere else on the banner - https://dashboard.radixdlt.com/network-staking/validator_rdx1s0sr7xsr286jwffkkcwz8ffnkjlhc7h594xk5gvamtr8xqxr23a99a

dcts:
@EvgeniiaVak we had no stake guide until a few hours back, so I thought better to link to the validator node. Now we have a thread so I changed the link as well. 

But you brought up a great idea IMO, we should definately have a staking guide in the gitbook documentation and link to that!

For now I suggest to keep the thread, its better than the radix dashboard link.

EvgeniiaVak (changes requested):
Now the banner has only the link to X. But it was very useful to have the link directly to the action needed as well. Can we limit the X link to the "Learn more about how to stake" button and leave the other banner area (or add another button) as a link to radix staking dashboard? Maybe in another PR, but if we only can have one link in the banner atm I vote for reverting to radix staking dashboard 😃

dcts:
> Can we limit the X link to the "Learn more about how to stake" button and leave the other banner area (or add another button) as a link to radix staking dashboard? 

While this is possible, I think having a single link for the banner makes more sense as its (1) less confusing for the user (its not something I would expect that different areas of the banner lead to different targets) and (2) its more scalable to work with future promotion banners.

dcts:
Reverted to radix dashboard link to get this PR merged :) @EvgeniiaVak

EvgeniiaVak (approved):

PR #351, opened 2024-04-16
Claims component improve state

dcts:
I suggest the following changes:
- all asyncThunk functions should return a target value, or throw an error (prevent just returning "undefined" as we won't be able to catch those errors)
- all state changes that are needed after any asyncThunk call should be handled inside the extraReducers (.fulfilled)
- all errors that happen during asynThink should be forwarded to the extraReducers as well and handled there (.rejected)
- Maybe I don't have the full context, but I believe deep-copying the full state with `JSON.parse(JSON.stringify(state))` is unneccessary (please correct me if I'm wrong).

I did adapt only `fetchAddresses` and `fetchReciepts` according to the above suggestions. I did not change `fetchAccountRewards` nor `fetchOrderRewards`. I can refactor those as well if that is useful @nguvictor 

I added some StateDebug component to show the state as a table on top of the page (instead of showing the header). This will help quickly debugging without opening conosle all the time.


src/app/state/rewardSlice.ts
```diff
@@ -32,8 +33,14 @@ interface RewardConfig {
   };
 }
 
-function findFieldByName(fieldName: string, fields: any[]) {
-  return fields.find((field) => field.field_name === fieldName);
+function findFieldValueByNameOrThrow(fieldName: string, fields: any[]): string {
```
dcts, line 36: This function should throw an error, as getting the component addresses is a mandatory step. Without it we won't be able to fetch rewards or claim, so better to throw error and handle the error accordingly.

src/app/state/rewardSlice.ts
```diff
@@ -83,53 +90,55 @@ type NonFungibleResource = NonFungibleResourcesCollectionItem & {
...
+>("rewards/fetchReciepts", async () => {
   const rdt = getRdt();
-  if (!rdt) return;
+  if (!rdt) {
+    throw new Error("RDT initialization failed");
+  }
+
+  // Question from @dcts: should those be really taken from .env
+  // or are we fetchnig those too with fetchAddresses ?
   const claimComponentAddress = process.env.NEXT_PUBLIC_CLAIM_COMPONENT;
```
dcts, line 106: Are those addresses supposed to be inside the .env or should those also be fetched by fetchAddresses?

src/app/state/rewardSlice.ts
```diff
@@ -83,53 +90,55 @@ type NonFungibleResource = NonFungibleResourcesCollectionItem & {
...
-
-    if (accountReceiptVault && accountReceiptVault?.vaults.items.length > 0) {
-      dispatch(
-        rewardSlice.actions.updateReciepts(
-          accountReceiptVault?.vaults.items[0].items as string[]
-        )
-      );
-    }
-  } catch (error) {
-    return undefined;
```
dcts, line 129: I suggest to let the error being thrown, catching the error to return undefined will make this error not catchable.

EvgeniiaVak:
@dcts the build failed with `Error: 'HeaderComponent' is defined but never used.`, there are also a bunch of `console.log` warnings. Does your IDE support ESlint highlighting? If you enable it you will see all warnings and errors directly while implementing the changes.

EvgeniiaVak:
> Maybe I don't have the full context, but I believe deep-copying the full state with JSON.parse(JSON.stringify(state)) is unneccessary (please correct me if I'm wrong).

@dcts I added copying because redux refused to work with the original state because it was not a plain object, so unless something changed inside AlphaDEX SDK, we either need to either stringify the whole object or extract only the fields we need. I should have left a comment about it though.

fliebenberg:
> @dcts the build failed with `Error: 'HeaderComponent' is defined but never used.`, there are also a bunch of `console.log` warnings. Does your IDE support ESlint highlighting? If you enable it you will see all warnings and errors directly while implementing the changes.

I think for now it is OK to leave the console.log lines in as this is only a merge with the claims-component branch and we are still actively debugging and optimizing the code. We will get rid of all these when we merge with main. 
Will we be able to merge the PR at all if it has warnings though?

fliebenberg (approved):
I am happy with all these changes.

EvgeniiaVak:
ah, it's not to the main branch, my bad, I don't mind console log warnings or build errors then 👍

EvgeniiaVak (approved):

dcts:
Thanks for all the feedback. I have eslint but didnt care that it doesnt build as its WIP.

dcts:
# Questions for @fliebenberg 

1. can you check whether the loaded data looks correct (https://github.com/DeXter-on-Radix/website/pull/351/commits/74bf7d64ace9881f03aa44452f9659c36464be1d)
2. can you provide the claims component nft address for mainnet? Currently I only have it for stokenet ()
3. I have now all receiptIds, they look like this (stokenet):
```
resource_tdx_2_1ng32esp95nmay5er9ycmgwwlf4yk8ef9kedfz3ap7myu0mxy3ayahv#78#
resource_tdx_2_1n2at3fhhq5vqqput5wz8tlutkfx42uyxcjhq9zcqypt3pxqw68drsl#358#
resource_tdx_2_1ng6vf9g4d30dw8h6h4t2t6e3mfxrhpw8d0n5dkpzh4xaqzqha57cd2#216#
resource_tdx_2_1ng6vf9g4d30dw8h6h4t2t6e3mfxrhpw8d0n5dkpzh4xaqzqha57cd2#217#
```
Does this look right? As I am not able to fetch orderRewards, Instead getting this error:
image

# General questions:
I found 2 naming issues that I find not intuitive (if there are no objections I suggest to change those):
- rewardSlice should be rewardsSlice IMO, as the page is called rewards (pluralized). I always type it with "s" and get an error.
- should the state not better be called `accountRewards` instead of `accountsRewards` (pluralized with "s")? Even thought the rewards are for multiple accounts (in the future), when looking at the datastructure it is NOT ambigous. To me pluralization of accounts feels counter intuitive.

# TODO's:
- [ ] IMO we should extract the accountRewards data that we need in a structure that we define, instead of storing the whole API response inside our state. When this is fixed we can also remove the deep copying of complex objects to prevent serialization error
- [ ] rename above issues if there are no objects
- [ ] add component nft address for mainnet
- [ ] rename receiptIds into receiptIdentifiers as those are now not only the Ids but the full identifiers

fliebenberg:
Answers/Comments to your post above...
1. Your hard-coded data looks right. I will write a function to fetch this from the API.
2. There is not a claims component for mainnet yet. We will launch one once we are ready to test on mainnet.
3. Your receipt ids look right, but I will have a look at why you are getting an error with the API call.

I agree that rewardsSlice sounds more intuitive, although it is slightly akward to type with the 2 ss next to each other. But dont have any objection to changing it.
I think we should keep the state as accountsRewards and it refers to the rewards for multiple accounts (even though we are only loading one account for now). It also distinguishes it from accountRewards which is the rewards for a single account. So e.g. you would probably iterate through the accountsRewards and get accountRewards in each iteration.

fliebenberg:
I am going to go ahead and merge this branch with the claims-component branch so that we have one branch we can contribute to.

fliebenberg:
Does this look right? As I am not able to fetch orderRewards, Instead getting this error (looks like getOrdersRewardsApiData returns empty entries):

I am not able to replicate the error you reported. Everything seems to work fine for me. If you are still experiencing the problem ,we can have a look at it live when you have a minute.

PR #353, opened 2024-04-17
Buy sell panel

dcts:


fliebenberg:
Hi @dcts You will need to provide a bit more info for this branch. Not sure what we are merging. It is also still failing some tests.

dcts:
> Hi @dcts You will need to provide a bit more info for this branch. Not sure what we are merging. It is also still failing some tests.

Hey @fliebenberg , please ignore for now as this is a draft. Will provide full details once finished :)

dcts:
@fliebenberg this PR is now ready to review! I finalized the following changes based on your comments:
1. max-width increased such that it will be fully filled on desktop
2. Increased price chart but kept the orderInput without scroll, as Mauro suggested
3. postonly is now inside the order input container
4. when user is not connected the button now says "Connect Wallet to Trade" and prevents the user from submitting trades. The quote is still being fetched, I think thats a nice feature to be able to "preview" your transaction even when not logged in.

EvgeniiaVak:
@dcts https://www.figma.com/file/Zjff03BX35fREatqzcWDS7/DeXter-UI?type=design&node-id=1507-1243&mode=design&t=iYmijyTOr4E0fxCg-0 this is the most recent design, right?

dcts:
> @dcts https://www.figma.com/file/Zjff03BX35fREatqzcWDS7/DeXter-UI?type=design&node-id=1507-1243&mode=design&t=iYmijyTOr4E0fxCg-0 this is the most recent design, right?

Yes, but there were some things that we improved and are not reflected in the design, I added comments.

src/app/state/orderInputSlice.ts
```diff
@@ -57,114 +60,160 @@ export interface TokenInput {
...
+  const isValidLimitOrder = type === OrderType.LIMIT && price > 0;
+  return isMarketOrder || isValidLimitOrder;
+}
+
+export function tokenIsSpecified(specifiedToken: SpecifiedToken): boolean {
+  return specifiedToken !== SpecifiedToken.UNSPECIFIED;
+}
 
 // for getting balances out of pairSelector slice
+// TODO(dcts): ask @chaotic whether this can live inside pairSelector now that
```
EvgeniiaVak, line 215: yes, sure, and probably it will live in the `appSlice` after Fred's refactoring of the state

src/app/utils.ts
```diff
@@ -347,6 +351,55 @@ export function detectBrowserLanguage(defaultLanguage: string = "en"): string {
...
+  if (formattedValue.endsWith(separator)) {
+    return formattedValue;
+  }
+  // Split and limit fraction scale as before
+  let [whole, fraction] = formattedValue.split(separator);
+  if (fraction && fraction.length > scale) {
+    fraction = fraction.substring(0, scale);
+  }
+  return fraction ? `${whole}${separator}${fraction}` : whole;
+}
```
EvgeniiaVak, line 401: it would be great to have unit tests for this function

EvgeniiaVak (approved):
Great work! The new order input is more beautiful and functional. 

Some minor style issues and tests can be done in separate PRs.

src/app/utils.ts
dcts: Done!

EvgeniiaVak (approved):

dcts:
Final testing was conducted yesterday and today. The following bugs were fixed:
- update balance after order is submitted (reported by [@davimeloag](https://t.me/Davimeloag))
- update balance after data sharing is updated
- when selecting MAX: reserve 3 XRD to pay transaction fees (reported and suggested by [@John101P](https://t.me/John101P))

EvgeniiaVak (changes requested):

src/app/utils.ts
```diff
@@ -347,6 +351,55 @@ export function detectBrowserLanguage(defaultLanguage: string = "en"): string {
...
+      MNI: 0,
+      OCI: 2,
+      PLANET: 0,
+      RDK: 0,
+      WEFT: 2,
+      XRAW: 0,
+      XUSDC: 2,
+    }[input.toUpperCase()] || 2
+  );
+}
```
fliebenberg, line 376: We need to make sure we replace this with a dynamic lookup. 
I am not sure how this is used, but there is already a max decimals value for each token in a pair in the pairInfo. I dont think 2 is enough for most use cases.

dcts: This is currently only used to display the precision in the "available: 1'000.00 XRD" label. 

But I agree we should figure out dynamic precision that automatically adapts when new tokens get added. I am tracking this issue here: https://github.com/DeXter-on-Radix/website/issues/371

src/app/components/OrderInput.tsx
```diff
@@ -520,6 +530,46 @@ function CurrencyInputGroupSettings(
     );
   };
 
+  const setPercentageAmountToken1 = (percentage: number, isXRD: boolean) => {
```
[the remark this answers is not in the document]
dcts: Added comments and created XRD_FEE_ALLOWANCE const as suggested by @fliebenberg. 

@EvgeniiaVak yes you're right we reserve 3 XRD whenever the user sets 100% on a currency input that is XRD, independant of the positioning (token 1 or token 2), e.g. it works for both DEXTR/XRD but also XRD/xUSDC pair.

fliebenberg (approved):

EvgeniiaVak (approved):

PR #355, opened 2024-04-19
Claims component dcts

dcts:
Added some features:
- refactor error handling
- refactor code duplication in rewardSlice
- add basic success UI (including new state `showSuccessUi`)
- refactor claimRewards function to be asnyc + hook into DexterToast.pending notification
- "no rewards to claim" usecase covered.

Still pending:
- make reward detail table prettier


fliebenberg:
I see the reward details are not shown in the latest commit anymore? Why are we excluding it?
When I use the page while not logged in to a wallet, it correctly shows me that I need to connect a wallet to see my rewards. When I then connect to a wallet, the page does not update. Also, I get 2 requests to connect to a wallet?

dcts:
@fliebenberg some comments:
- I excluded details page as I expected us to launch to the public, and IMO its better to launch without it rather than having a UI thats not very pretty. Now that I am back I will implement the UI during the next ~2 days
- the issues described with wallet connection I sometimes too have, sometimes it works. But its hard to replicate. I suspect that it has something to do with the messy state that we'll hopefully fix during our call tomorrow. We initialize the redux store on both pages (trade and rewards) so I can imagine duplicate login requests might happen due to that. 

I suggest to keep this PR open until I implement the details UI, unless someone needs to work on it.

dcts:
Added rewards page details table ui:

image

fliebenberg:
Thanks. I am going to merge this with the claims-component branch now and will test with that branch.

PR #312, opened 2024-03-23
Claims component

nguvictor:

PR #378, opened 2024-05-03
store public env vars in repo

EvgeniiaVak:
In #312 a new env var was added with no mention of it in `CONTRIBUTING.md`, also we forgot to add it to deployment. This PR is aimed to make it easier to not forget adding the variables by always keeping publicly available variables in the repo itself.


EvgeniiaVak:
@fliebenberg Is there any specific reason [to fetch the addresses](https://github.com/DeXter-on-Radix/website/blob/e8689b75050474eeb3a6ea3eb8817f4ca851de1c/src/app/state/rewardSlice.ts#L285-L286) instead of hardcoding?

fliebenberg:
All the fetched addresses are related to the claim component. So every time you change the claim component address you HAVE to also update all the other addresses and they are not easy to get unless you read the component state. So it makes more sense to only provide the new component address and then let the app read the other addresses. It means all the other addresses will always be correct for the current component, whereas if they are all set manually you get get situations where some still refer to an old component which will lead to unexpected values.

nguvictor (approved):

fliebenberg (approved):

PR #382, opened 2024-05-06
New state management principles

fliebenberg:
I have created this PR to discuss the principles we want to follow when managing state in our app.
As a start I suggest you read the STATE.md file in the /state folder. It lists some of the priciples I suggest we use for state management.
As an example of these principles I created a "user" state folder that follows the principles I suggested as an example of how this implementation could look. All other state was left untouched for now.
I will probably set up a meeting for us to discuss this example and get everyone's feedback. For now please feel free to comment and give suggestions of improvements/changes.


src/app/state/STATE.MD
```diff
@@ -0,0 +1,86 @@
...
+    initialState,
+    reducers: {
+      ...
+    },
+    extraReducers: (builder) => {
+      ...
+    }
+  })
+  ```
+- an export of the available actions for the slice called ```<slice_name>Actions. The purpose of this export is to only expose the actions you want users to call (which might exclude certain reducers and/or thunks that should only be used by the slice internally).
```
dcts, line 66: Since this example slice is `userSlice` the actions are userActions. But in general this could be called `rewardActions` or `radixActions` and does not neccessarily represent user actions in all cases.

```suggestion
- an export of the available actions for the slice called ```<slice_name>Actions. The purpose of this export is to list all actions that can be called from outside of the slice (which might exclude certain reducers and/or thunks that should only be used by the slice internally).
```

src/app/state/STATE.MD
```diff
@@ -0,0 +1,86 @@
...
+  ```
+- The initial state of the slice stored in a constant called `initialState`
+  ```
+  const initialState: userState = {
+    ...
+  }
+  ```
+- The createSlice function for the slice called `<slice_name>Slice`
+  ```
+  export const unsrSlice = createSlice({
```
dcts, line 55:
```suggestion
  export const userSlice = createSlice({
```

src/app/state/accountHistorySlice.ts
```diff
@@ -26,11 +26,13 @@ export interface AccountHistoryState {
 const initialState: AccountHistoryState = {
   trades: [],
   orderHistory: [],
+  // these tables are component specific and should not be in app state
```
dcts, line 29: Makes sense. But to ensure I get the idea right: the underlying data that is needed to derive the tables (all user trades) should be in global user state, right?

src/app/state/accountHistorySlice.ts
```diff
@@ -26,11 +26,13 @@ export interface AccountHistoryState {
...
   selectedTable: Tables.OPEN_ORDERS,
   tables: Object.values(Tables),
 };
 
 // ASYNC THUNKS
+// thunks should change state or call other functions that change state. This is just a normal async function that fetches data.
```
dcts, line 35: Can you explain why account history fecthing does not change the state? As far as I understand those trades done by the user should be stored in global state (effectively updating state once the data is fetched).

src/app/state/accountHistorySlice.ts
```diff
@@ -53,6 +55,7 @@ export const fetchAccountHistory = createAsyncThunk<
   }
 });
 
+// this is just a normal async function and should not be a thunk.
```
dcts, line 58: This function currently uses `thunkAPI.dispatch` and `thunkAPI.getState()` so it looks like its interacting with state. I'd need again an explenation why it should not be a thunk.

dcts (commented):
# Question1: Updating Packages and Engine

I'm wondering whether we should upgrade the packages 
```
 "@radixdlt/radix-dapp-toolkit": "^1.4.4",
 "alphadex-sdk-js": "^0.12.14",
```
And also the engine
```
"target": "es2015",
```
in a separate PR to test in isolation and then merge main into this PR?

# Question2: local state

This PR suggest to use local state for component wide state and redux for global state. I would suggest to make it even more specific, here's my suggestion:
- **`createSlice()`**: global state as described in this PR.
- **`useContext`**: page wide local state that can be shared among different components on a single page. 
- **`useState`**: if we have atomic components (reusable UI components like buttons or headings etc), those should not depend on either global nor local page wide state. 

# Note: use AsyncThunk only for state interactions

the async thunk API contains of the following utilities:
- `asyncThunk.getState()` 
- `asyncThunk.dispatch()` 
I've previously ran into the issue where the action at the end of an asyncThunk was not defined by extraReducers but instead the asyncThunk in the end just dispatched an action (here is the reference: https://github.com/DeXter-on-Radix/website/pull/351/files#diff-64e03be502081f4819f8befa5d2e088c3f729ebb3906e3e3318190e9477a9b8a) 

Maybe it would be good to talk through these use cases quickly, as we could handle state updates either by extraReducers or by dispatching actions at the end of the function.

Also I would be curious to talk through the 2 comments where you mentioned "this does not need to be an async thunk" (inside the AccountHistorySlice).

src/app/state/accountHistorySlice.ts
fliebenberg: Yes. The general data should be stored globally so it is available to many components. The formatting of the data for the specific use in the component must be done locally.
It means the app can keep the general data updated in one place (state) and these changes can be fed to the various components automatically (by using slices).

src/app/state/accountHistorySlice.ts
fliebenberg: Your right. When I was adding these coomments I must have missed that there is an associated extraReducer for this thunk that does change state.

src/app/state/accountHistorySlice.ts
fliebenberg: Yep, you're right again. The fact that it uses state does not make it a thunk. Any component/code can use state. But, the fact that it dispatches an action as a result of something else happening (the transaction has been submitted) does make it a valid thunk.

fliebenberg:
Thanks for all the comments @dcts . 
Question 1 and 4: I agree we should roll out some of these changes in smaller PRs to be able to incrementally update main. So far I have tried not to change any existing state and have just been adding extra state. 
Question 2: I like this suggestion. Smaller (view only/atomic) components should preferably get all their state as props from their parent and feed actions back to their parents. That helps to make them truly independent (they need no knowledge of state structure etc).
Question 3: It would be useful to get a clear definition of when something is a Thunk or not and where to deal with dispatching actions (in the thunk or extraReducers). Maybe a chat about this will be easiest once others have had a chance to look at eh suggestions as well. BTW, you were right that both the functions I said were not true thunks are actually thunks. I corrected myself by removing my comments. :-)
Thanks again for your feedback.

Issue #371, opened 2024-04-30
Implement decimals depending on coins


dcts:
Currently we show 8 decimals for most coins, and we round by 6 decimals in the quoteDescription. Ideally we would show decimals for price / amount based on the coin price. 

e.g. we should show higher precision for coins like FLOOP (as its very expensive) compared to XRD.

Reported by John101:

![image](https://github.com/DeXter-on-Radix/website/assets/44790691/87cea989-6ac7-4c8c-a562-3b82d7f8ffbf)

![image](https://github.com/DeXter-on-Radix/website/assets/44790691/10882cf8-3079-4b5d-8601-7d188386fc39)

EvgeniiaVak:
@fliebenberg could you point what SDK properties can we use here if any? I've seen these [properties in the docs](https://www.npmjs.com/package/alphadex-sdk-js): 
<img width="600" alt="Screenshot 2024-05-06 at 16 41 31" src="https://github.com/DeXter-on-Radix/website/assets/27793901/197f74c3-0573-4ad1-9bf1-cc3b44461b03">

But I guess precision is not the same as the number of decimals allowed.

PR #384, opened 2024-05-06
Fix precision issue

### LINKED ISSUE #383: Specifying MAX for non-XRD inputs lead to insufficient funds
When we add max on a non-XRD input field with high precision, we get an "insufficient balance" error inside the wallet (with approximately 50% probability). This is most likely caused by `adex.createExchangeOrderTx`.

# To Reproduce
1. buy some test DEXTR tokens (on stokenet)
2. ensure you have a high decimal amount in your balance, if you want to reproduce, ensure you have exactly `1238.5599423734525 DEXTR` in your wallet (you can send the difference to another wallet)
3. Go to DEXTR/XRD -> SELL -> specify MAX by clicking the `Available: 1238.55` label
4. Send transaction
5. This error should apper:
image

We are sending the full balance to `adex.createExchangeOrderTx` but for some reason the amount is then rounded in a weird way to `...38` instead of `...37`. 

To fix this we can simply truncate the amount we send to `adex.createExchangeOrderTx` to the 8th decimal (truncate, NOT round).

dcts:
This PR:
- fixes #383 
- replaces rounding with truncating on all balances shown in the frontend, to prevent situations like these:
image

### DIFF
--- a/src/app/components/OrderInput.tsx
+++ b/src/app/components/OrderInput.tsx
@@ -544,10 +544,11 @@ function CurrencyInputGroupSettings(
       isXRD ? balanceToken1 - XRD_FEE_ALLOWANCE : balanceToken1,
       Calculator.divide(Calculator.multiply(balanceToken1, percentage), 100)
     );
-    const targetAmountRounded = truncateWithPrecision(targetAmount, 8);
+    // TODO: revert truncating to 8th decimal once adex has fixed adex.createExchangeOrderTx bug
+    const targetAmountTruncated = truncateWithPrecision(targetAmount, 8);
     dispatch(
       orderInputSlice.actions.setTokenAmount({
-        amount: targetAmountRounded,
+        amount: targetAmountTruncated,
         bestBuy,
         bestSell,
         balanceToken1: balanceToken1,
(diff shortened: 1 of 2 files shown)


EvgeniiaVak (changes requested):
There is too much variation in how many decimal places are shown:
Screenshot 2024-05-06 at 14 38 35
photo_2024-05-06 14 43 10

Especially in the onhover info, I would expect there to be the same number as [in the transaction](https://stokenet-dashboard.radixdlt.com/transaction/txid_tdx_2_1czvc0l4gvfqtx7lfhmmg8fyk8l5rn7pzd0jdsj33d6qn66fmr55sk6mgl5/summary) (in this case it was 3 decimal places).

dcts:
@EvgeniiaVak updated the frontend precision to be the same for both the label and the description. I am against setting the precision to be 3 as specified in the wallet app because we run into multiple issues:
1. we don't control the wallet and don't know how they compute the precision. Will they also show only 3 decimals on high value coins like FLOOP? What if they decide to update the app? It will be tedious and less maintainable. Better to have our own logic that is consistent (and I agree we should be showing values formatted the same on our end).
2. I am against 3 decimals, because it can easily be confused as thousands delimitter. 
3. Our logic to show precision is currently hardcoded [here](https://github.com/DeXter-on-Radix/website/blob/main/src/app/utils.ts#L356-L376) but should be adapted to be more resiliant and dynamically computed, this is another bug (#371) and is not part of this PR. 

Given the above I just changed the display to be 2 decimals for both frontend displays, hope that is fine for now to merge this PR.

EvgeniiaVak (approved):
Approving as these changes fixes a major issue, my comment above can be dealt with in a separate PR.

nguvictor (approved):

PR #386, opened 2024-05-06
Changes to subscriptions file

fliebenberg:
This PR is to make a few suggested changes to the subscriptions file.

The biggest change is that it now exports a variable called rdt, which can just be imported and used anywhere else in the app. The rdt variable is initialised in the init subscriptions function which is called once when the app loads from a useEffect in the root layout component.
I also simplified the adex init call and added an interval to the init subscriptions function to show where to create intervals that need to always run in the app. This interval will be used in the new state management.

These changes were part of the new state management changes, but I wanted to do a separate PR so we can discuss and approve these changes separately.

### DIFF
--- a/src/app/subscriptions.ts
+++ b/src/app/subscriptions.ts
@@ -17,10 +17,10 @@ import { AppStore } from "./state/store";
 import { rewardSlice } from "./state/rewardSlice";
 
 export type RDT = ReturnType<typeof RadixDappToolkit>;
+export let rdt: RDT;
 
-let rdtInstance: null | RDT = null;
 export function getRdt() {
-  return rdtInstance;
+  return rdt;
 }
 
 export function getRdtOrThrow() {
@@ -30,33 +30,23 @@ export function getRdtOrThrow() {
   }
   return rdt;
 }
-function setRdt(rdt: RDT) {
-  rdtInstance = rdt;
-}
 
 let subs: Subscription[] = [];
 
 export function initializeSubscriptions(store: AppStore) {
-  let networkId;
-  switch (process.env.NEXT_PUBLIC_NETWORK!) {
-    case "mainnet":
-      networkId = RadixNetwork.Mainnet;
-      break;
-    case "stokenet":
-      networkId = RadixNetwork.Stokenet;
-      break;
-    default:
-      networkId = RadixNetwork.Stokenet;
-  }
-  rdtInstance = RadixDappToolkit({
-    dAppDefinitionAddress: process.env.NEXT_PUBLIC_DAPP_DEFINITION_ADDRESS!,
-    networkId,
+  rdt = RadixDappToolkit({
+    dAppDefinitionAddress:
+      process.env.NEXT_PUBLIC_DAPP_DEFINITION_ADDRESS || "",
+    networkId:
+      process.env.NEXT_PUBLIC_NETWORK == "mainnet"
+        ? RadixNetwork.Mainnet
+        : RadixNetwork.Stokenet,
   });
-  rdtInstance.walletApi.setRequestData(
-    DataRequestBuilder.accounts().exactly(1)
-  );
+  // TODO: "black" on the light theme
+  rdt.buttonApi.setTheme("white");
+  rdt.walletApi.setRequestData(DataRequestBuilder.accounts().exactly(1));
   subs.push(
-    rdtInstance.walletApi.walletData$.subscribe((walletData: WalletData) => {
+    rdt.walletApi.walletData$.subscribe((walletData: WalletData) => {
       const data: WalletData = JSON.parse(JSON.stringify(walletData));
       store.dispatch(radixSlice.actions.setWalletData(data));
 
@@ -64,9 +54,6 @@ export function initializeSubscriptions(store: AppStore) {
       store.dispatch(fetchBalances());
     })
   );
-  setRdt(rdtInstance);
-  // TODO: "black" on the light theme
-  rdtInstance.buttonApi.setTheme("white");
   let network;
   switch (process.env.NEXT_PUBLIC_NETWORK!) {
     case "mainnet":
--- a/src/app/subscriptions.ts
+++ b/src/app/subscriptions.ts
@@ -19,24 +19,12 @@ import { rewardSlice } from "./state/rewardSlice";
 export type RDT = ReturnType<typeof RadixDappToolkit>;
 export let rdt: RDT;
 
-export function getRdt() {
-  return rdt;
-}
-
-export function getRdtOrThrow() {
-  const rdt = getRdt();
-  if (!rdt) {
-    throw new Error("RDT initialization failed");
-  }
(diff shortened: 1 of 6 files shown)


dcts (dismissed):
Looks way cleaner IMO, I like it.

Also tested both main user journies (buy tokens + claim rewards) on stokenet and everything worked as expected.

dcts:
Deployment failed with this logs (I could not retrigger deployment, we should try again maybe later and if the same error persists contact support, looks like a cloudflare outage from the logs as I was able to run this locally @fliebenberg):

```
21:42:41.448 | Finished
-- | --
21:42:41.449 | Note: No functions dir at /functions found. Skipping.
21:42:41.449 | Validating asset output directory
21:42:42.106 | Deploying your site to Cloudflare's global network...
21:42:48.602 | Failed: an internal error occurred. If this continues, contact support: https://cfl.re/3WgEyrH
21:42:48.602 | Error: Failed to publish assets. For support, join our Discord (https://discord.gg/cloudflaredev) or create a ticket and reference the deployment ID: 00850e30-d5e4-43bd-afa0-661044cd50f6
```

https://dash.cloudflare.com/0c61b59baf9dcb09b3d02e0bb4f998a1/pages/view/dexter/00850e30-d5e4-43bd-afa0-661044cd50f6

src/app/subscriptions.ts
```diff
@@ -17,69 +17,35 @@ import { AppStore } from "./state/store";
...
-      break;
-    default:
-      networkId = RadixNetwork.Stokenet;
-  }
-  rdtInstance = RadixDappToolkit({
+  rdt = RadixDappToolkit({
     dAppDefinitionAddress: process.env.NEXT_PUBLIC_DAPP_DEFINITION_ADDRESS!,
-    networkId,
+    networkId:
+      process.env.NEXT_PUBLIC_NETWORK == "mainnet"
```
EvgeniiaVak, line 29:
```suggestion
      process.env.NEXT_PUBLIC_NETWORK! == "mainnet"
```
This non-null TS assertion just ensures we have more readable errors if the env var is not set up.

src/app/subscriptions.ts
```diff
@@ -17,69 +17,35 @@ import { AppStore } from "./state/store";
...
-  const rdt = getRdt();
-  if (!rdt) {
-    throw new Error("RDT initialization failed");
-  }
-  return rdt;
-}
-function setRdt(rdt: RDT) {
-  rdtInstance = rdt;
-}
+export let rdt: RDT;
```
EvgeniiaVak, line 20: I'm surprised TS allows you to do that without an error. You are declaring `rdt` as RDT type always, but not assigning a value immediately. That means at some point in time during the application init this variable is `undefined` and we need to not forget to deal with such cases accordingly everywhere we are using the rdt instance. 

```suggestion
export let rdt: RDT | undefined;
```

EvgeniiaVak (changes requested):

src/app/subscriptions.ts
fliebenberg: Yep, agree. Must have missed that one.

dcts (approved):

src/app/subscriptions.ts
EvgeniiaVak: @fliebenberg what about this one? Are you sure rdt will always be initialized before usage? Even on slow CPU clients?

nguvictor: I'm thinking it might be good idea that we can trust rdt to be always initialized, otherwise my opinion is that the app should fail as it's an important dependency for many different slices in the application. Also, as far as I know we wouldn't want control of initializing rdt to fall on random slices.

dcts: GPT4 suggest adding `--strictNullChecks`:

"TypeScript allows this behavior because the variable's type (RDT) does not exclude undefined or null by default. If you want TypeScript to enforce that a variable always has a value of its specific type from the moment it is defined, you can configure TypeScript to use strict null checks (--strictNullChecks option)"

EvgeniiaVak (changes requested):

src/app/subscriptions.ts
fliebenberg: I agree that it should be fine for the app to crash (essentially it wont start) if rdt is not initialised. rdt is initialised with the RadixDappToolkit function which is a synchronous function that returns an instance of the RadixDappToolkit type. So it will be completely initialized or fail (e.g. if some of the settings in the code is incorrect) before the code can go on to load the rest of the app.

EvgeniiaVak: We probably need some global state variable that watches if the app is initialized and only render all the child components when the init script has finished.

fliebenberg: The best would be if we did not have to run the initSubscriptions in an effect at all. It is not an effect, it is just normal synchronous code that needs to run once when the app starts. Unfortunately I dont know Next well enough to know where that is.

fliebenberg: When I try to move the initSubscriptions function out of the useEffect I get this build error:

Error occurred prerendering page "/_not-found". Read more: https://nextjs.org/docs/messages/prerender-error
ReferenceError: document is not defined
    at oz (C:\Fred\Coding\dexter\website\.next\server\chunks\511.js:52:79749)
    at lA (C:\Fred\Coding\dexter\website\.next\server\chunks\511.js:52:165524)
    at x (C:\Fred\Coding\dexter\website\.next\server\chunks\361.js:2:13269)
    at V (C:\Fred\Coding\dexter\website\.next\server\chunks\361.js:1:8863)
    at nT (C:\Fred\Coding\dexter\website\node_modules\next\dist\compiled\next-server\app-page.runtime.prod.js:12:46138)
    at nT (C:\Fred\Coding\dexter\website\node_modules\next\dist\compiled\next-server\app-page.runtime.prod.js:12:59937)
    at nO (C:\Fred\Coding\dexter\website\node_modules\next\dist\compiled\next-server\app-page.runtime.prod.js:12:62795)
    at nI (C:\Fred\Coding\dexter\website\node_modules\next\dist\compiled\next-server\app-page.runtime.prod.js:12:65445)
    at nj (C:\Fred\Coding\dexter\website\node_modules\next\dist\compiled\next-server\app-page.runtime.prod.js:12:64852)
    at nO (C:\Fred\Coding\dexter\website\node_modules\next\dist\compiled\next-server\app-page.runtime.prod.js:12:63041)

EvgeniiaVak: This error happens because you need to make sure the init runs only in browser (rdt doesn't work otherwise), in react `useEffect` runs only in browser, maybe you can find some other way, but I think having a state variable is the simplest: 
- create a local variable with useState in layout component (something like isInitilized), set it to false
- after init finishes successfully set this variable to true
- render children only when isInitilized is true

dcts:
Closing as this is WIP. Please re-open if needed @fliebenberg :)

Issue #395, opened 2024-05-10
Make promo banner a carousel


dcts:
To enable us to show multiple banners, we should make the banner a carousel that autoamtically changes the content on a set period (e.g. every 10 seconds). This will allow us to promote multiple things in parallel.

Here is a design mock ([figma file link](https://www.figma.com/design/Zjff03BX35fREatqzcWDS7/DeXter-UI?node-id=825-6498&t=XoNyy5lkJLRQBFmX-0
)):
<img width="1601" alt="Bildschirmfoto 2024-05-10 um 14 55 13" src="https://github.com/DeXter-on-Radix/website/assets/44790691/88297358-6938-4bf5-a389-c5ca84e35592">

I suggest to either code the carousel from scratch or use a small library to prevent loadtime increase too drastically.

dcts:
Issue fixed by @restylianos, PR #408 pending. Closing to prevent duplicate assignment.

PR #408, opened 2024-05-15
[FEATURE] promotion carousel

restylianos:
@dcts Hi, I created a carousel component based on the pre existing banner image. I thought that we should keep the 80px height available for > 600px screens  and allow things to "stretch" in smaller screens. Moreover, i avoided turning the promo banner to a carousel in order to allow all kinds of renders inside the banner appart from images. Please let me know if this approach makes sense.

Related to https://github.com/DeXter-on-Radix/website/issues/395


dcts (changes requested):
Wow this is looking great and works as expected, well done! 🔥💯

May I request some structural changes: Since the `Carousel` component depends on the height of `PromoBanner`, can we keep those 2 components inside a single file? If the usecase arises in the future, we can make the Carousel a standalone copmonent for sure, but at this moment I think keeping it together makes more sense.

Ideally we should merge it into a PromoBannerCarousel component that is initialized like this (whether you'd like to modify PromoBanner or create a wrapper for it is up to you):
```jsx
<PromoBannerCarousel
  interval={10000}
  items={[
    {
      imageUrl: "/promo-banners/validator-node-staking/desktop-600x80.svg",
      imageUrlMobile: "/promo-banners/validator-node-staking/mobile-600x200.svg",
      redirectUrl: "https://dashboard.radixdlt.com/network-staking/validator_rdx1s0sr7xsr286jwffkkcwz8ffnkjlhc7h594xk5gvamtr8xqxr23a99a",
    },
    {
      imageUrl: "/promo-banners/boilerplate/desktop-600x80.svg",
      imageUrlMobile: "/promo-banners/boilerplate/mobile-600x200.svg",
      redirectUrl: "",
    },
  ]}
/>
```

The logic should stay the same as you have implemented it:
- default of 10s interval
- only render carousel if items.length>1
- items should be an array of [PromoBannerProps](https://github.com/DeXter-on-Radix/website/blob/main/src/app/components/PromoBanner.tsx#L4-L8)

That way we keep all logic in one place and keep intialization also clean.

dcts (changes requested):
Cool, thanks for the changes, although I think my last review might have been misleading. I hope the comments in this review clarify the idea better. I want to avoid splitting the content into multiple files, as the components depends on each other. 

Also I'd like to rename the component to PromoBannerCarousel.

So the general idea is to put all content into `PromoBannerCarousel.tsx` and have a single component `PromoBannerCarousel` that is exported.

src/app/components/PromoBannerCarousel.tsx
```diff
@@ -0,0 +1,115 @@
+import React, { useCallback, useEffect, useMemo, useState } from "react";
+
+export interface PromoBannerProps {
+  imageUrl: string; // 600x80
+  imageUrlMobile: string; // 600x200
+  redirectUrl: string; // target redirect address when banner is clicked
+}
+
+interface CarouselProps {
```
dcts, line 9:
```suggestion
interface PromoBannerCarouselProps {
```

src/app/components/PromoBannerCarousel.tsx
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// The banner display logic is different depending on screensize:
+// < 600px  : show mobile banner (600x200), and scale down linearly
+// > 600px  : show desktop banner (600x80) without scaling, center content
+
+export function PromoBannerCarousel({
+  items,
+  interval = 10000,
+}: CarouselProps) {
```
dcts, line 21:
```suggestion
}: PromoBannerCarouselProps) {
```

dcts (dismissed):

dcts:
Looks great! Thanks again!

dcts (approved):

restylianos:
Closing since a copy was made on the original repo

PR #417, opened 2024-05-22
[BUGFIX] remove invalid coin url

dcts:
There are a few pairs on DeXter that are from very unknown projects. I am wondering whether it would be better to remove them. At least we should have a minimal requirement that the token needs to have an image (which 3TR for example does not). 

This PR removes the following pairs:
```js
const UNVERIFIED_PAIRS = [
  "component_rdx1cqzv2v63gaexzf28g8slvr7y0xtdhzjetckqj0n0vwk9syk3klxtv2", // 3TR/XRD
  "component_rdx1crzqxsshnx4mk2474vy9ans6p8v0njph6hvfrt025c7uq8lwf9h0w9", // I£/XRD
];
```


nguvictor:
What would you think about doing a verified list vs unverified list? That way we would be able to have high control over reacting to every new listing. It would be safer for our users and will end up with fewer updates needed overall.

nguvictor (dismissed):

dcts:
@nguvictor yes I personally would prefer to have coins go through a formal "application" and we add them purposefully to a verified list, but upon discussion on TG this would need to be approved by the community first (hence we would need to draft a formal proposal and put up for a DAO vote). For now we can resolve this issue simply by sorting the pairs by by trade volume and de-emphasize coins with low volume, which added as issue in #419. 

Here is the start of the TG discussion for reference:
https://t.me/dexter_discussion/16242

But I would argue we should remove `3TR/XRD` as it has an invalid iconUrl that leads to the page making http requests to a weird looking website (blessings.org.uk). So I adapted the PR to only remove that one for now.

image

nguvictor (dismissed):

fliebenberg:
This is completely a Dexter community decision. The only way it will work is to either have a blacklist (list of pairs we specifically hide) or a whitelist (list of pairs that we display). Of these 2 options I prefer the blacklist method as it means we automatically show all pairs listed on AlphaDEX, unless we specifically decide not to list a pair.
In this PR you are also essentially suggesting a blacklist which I am happy with, but I do think we just need to confirm with the community that we will be doing this. We can then also ask them to suggest other (future) pairs they might want to blacklist/remove from Dexter.
Instead of calling the variable "UNVERIFIED_PAIRS" I would rather call it "BLACKLISTED_PAIRS". In future there might be a lot of pairs added that was not specifically added/verified by Dexter but that we are happy to include. So not all pairs that we show might be "verified" by Dexter.

dcts:
Ok, just started working on a formal proposal on how DeXter should handle the list of tokens (together with John101). 

For now I am just concerned that the request to blessings.org.uk looks suspicions. I did the following changes:
- renamed unverifiedPairs to blacklistedPairs
- blacklistedPairs is an empty array for now, so no pairs are blacklisted currently, as this requires formal community approval (added a comment in the code as well)
- I swapped the token URL for 3TR to prevent suspicions looking link showing as error in the console

nguvictor (approved):

nguvictor (dismissed):

nguvictor (approved):

PR #418, opened 2024-05-23
Promotion carousel

restylianos:
previous discussion was here: [408](https://github.com/DeXter-on-Radix/website/pull/408)


dcts:
Deploy preview is here:
https://ba894aff.dexter-1we.pages.dev/

I am not sure why it didn't get autotriggered. But in any case, theres nothing to see in the deploy preview, as there is still only a single image in the array. This PR just sets us up to add any amount of banners with auto-slide functionality (every 10 seconds).

dcts (approved):

nguvictor (approved):

PR #391, opened 2024-05-08
[UI enhancment] add sticky navbar in mobile views

restylianos:
relates to #388

PR #420, opened 2024-05-23
add sticky navbar on mobile views

restylianos:
Related to closed issue https://github.com/DeXter-on-Radix/website/pull/391

### DIFF
--- a/src/app/components/NavBar.tsx
+++ b/src/app/components/NavBar.tsx
@@ -33,7 +33,14 @@ const NavItems: { path: string; title: string }[] = [
 
 export function Navbar() {
   return (
-    <nav className="flex items-center justify-between w-full !h-[74px] !min-h-[74px] z-[100]">
+    <nav
+      className={
+        "sticky top-0 sm:border-b-0 border-b-2 sm:border-none border-gray-800 sm:static " +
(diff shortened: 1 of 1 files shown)


src/app/components/NavBar.tsx
```diff
@@ -33,7 +33,14 @@ const NavItems: { path: string; title: string }[] = [
 
 export function Navbar() {
   return (
-    <nav className="flex items-center justify-between w-full !h-[74px] !min-h-[74px] z-[100]">
+    <nav
+      className={
+        "sticky top-0 sm:border-b-0 border-b-2 sm:border-none border-gray-800 sm:static " +
+        "bg-base-200 flex items-center justify-between w-full !h-[74px] !min-h-[74px] " +
```
dcts, line 39: I just realized that the navbar should be transparent when on desktop and on the landing page, this looks a bit weird:

Bildschirmfoto 2024-05-23 um 22 40 38

I think to solve it it is sufficient to only add the background color for mobile screens (<640px). I think this should fix the issue:

```suggestion
        "bg-base-200 sm:bg-transparent flex items-center justify-between w-full !h-[74px] !min-h-[74px] " +
```

dcts (changes requested):

dcts (approved):

Radstakes (approved):

Issue #393, opened 2024-05-10
mobile responsive search suggestion overflow


SandipGyawali:
![image](https://github.com/DeXter-on-Radix/website/assets/66510112/ba56f038-d023-4a57-8eeb-a208fb7d48ad)

![image](https://github.com/DeXter-on-Radix/website/assets/66510112/f0f4133a-8aa7-4d00-9426-02709de57041)

search bar suggestion list overflow from it's maximum height.

SandipGyawali:
@EvgeniiaVak i guess this issue has not be addressed previously right ?

restylianos:
![Uploading Screenshot 2024-05-17 at 6.19.06 PM.png…]()

PR #421, opened 2024-05-23
fix suggestions overflow on mobile screens

restylianos:
Resolves [393](https://github.com/DeXter-on-Radix/website/issues/393)


dcts:
For other reviewers: 

fixes #393

# How to test

- go to deploy preview
- select screensize to be mobile
- open pair selector: only 3 items will show
- select any <li> item and click "edit as HTML", then copy the full HTML of the element and paste it 10 times (now we have 13 items, so we can test the bugfix).

nguvictor (approved):

Issue #268, opened 2024-02-20
Pair selector selects wrong pairs on mobile


dcts:
On mobile, the pair selector sometimes selects the wrong pair. 

To reproduce:
- open dexteronradix in the browser, open devtools, chose mobile view (alternatively, use your phone)
- click pair selector
- select FLOOP - XRD
- the pair DCG - XRD is selected instead. 

It cannot be a misclick due to narrow elements, as `DCG - XRD` pair is not even a neighbor of `FLOOP - XRD` pair. It looks like it just selects the first pair. But it doesnt do that consistently. As a rough estimate, selecting a pair on mobile only works in about 50% of the tries.

![bug](https://github.com/DeXter-on-Radix/website/assets/44790691/47f2b379-bca2-456d-a0e2-52ed802fba07)

nguvictor:
I was able to reproduce this as well. 
![mobile](https://github.com/DeXter-on-Radix/website/assets/3978285/7570d3a7-207f-401b-96ab-73c7a8cca72c)

PR #426, opened 2024-05-26
[Fix]: Wrong pair selector values

restylianos:
Resolves [268](https://github.com/DeXter-on-Radix/website/issues/268)

### DIFF
--- a/src/app/components/PairSelector.tsx
+++ b/src/app/components/PairSelector.tsx
@@ -242,7 +242,11 @@ export function PairSelector() {
                 onClick={() => {
                   selectOption(index);
                 }}
-                className={"!px-3 py-2 cursor-pointer hover:bg-base-300"}
+                className={`!px-3 py-2 cursor-pointer  ${
+                  highlightedIndex === index
+                    ? "bg-base-300"
+                    : "hover:bg-base-200"
+                }`}
                 style={{ marginTop: 0, marginBottom: 0 }}
                 key={`${id}-${index}`}
               >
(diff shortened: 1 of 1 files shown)


dcts:
Deploy Preview is here:
https://restylianos-pair-selector-sy.dexter-1we.pages.dev/

When I try it on testnet on mobile, the issue still occures. For example, in this video I open the pair selector and then select Adex/xrd but it doesnt work. Then I select cuppa/xrd next but it changes to adex/xrd.


https://github.com/DeXter-on-Radix/website/assets/44790691/88e1d180-820a-44ac-901e-2fecdff4ffec

dcts:
Awesome, this fixes the issue for me! Well done!
🎉🔥

Will review the code when I get home but its great to see this bug resolved, especially as radix wallet will soon be available for mobile!

restylianos:
The issue was that the onmouseenter event is not triggered in the same way across mobile and desktop devices. I would expect to have a never working filtering selector on mobile devices but it seems that the first time that an option is selected the onhover event is triggered and then the click event which is the same as the ??mouseenterevent?? (I am not sure if this is true since i cannot find a specific article which explains what happens in this scenario. I will update my comment if I manage to find anything relevant to this specific case). Based on the previous code version the onMouseEnter event was used on each option in order to change the highlighted index. After that, the selectOption function was relying on highlighted index in order to "perform" an element selection which was leading to the selection of a wrong element `const option = filteredOptions[highlightedIndex]`.

nguvictor:
I've had this intermittently happen on desktop as well, thanks for the fix!

nguvictor (approved):

restylianos:
> I've had this intermittently happen on desktop as well, thanks for the fix!

@dcts This was due to the useEffect sync issue

dcts (approved):

PR #440, opened 2024-06-01
customized order transaction submission

### LINKED ISSUE #425: Set transaction message to "Swap Using DeXter"
Similar to how other DEXes specify a message in the TX, we should add `"Market Order Using DeXter"` or `"Limit Order Using DeXter"`. It would have the following benefits:
- it adds **free exposure** to Dexter in the radix dashboard, people will start seeing these messages and might learn about DeXter
- it is **required for TokenTrek** dmany campaign, to be able to automatically verify a given action

Usually you can set a transaction message in the RDT API like this:

```js
const result = await rdt?.walletApi.sendTransaction({
 transactionManifest,
 version: 1,
 message
});
```

To be able to implement this for DeXtr this would require an adaptation of the Alphadex API (I believe, @fliebenberg). This is the adex API endpoint we are using to create orders:
- [`submitTransaction`](https://www.npmjs.com/package/alphadex-sdk-js#submittransaction-async-function-back-to-functions-list)
- We use this API here: https://github.com/DeXter-on-Radix/website/blob/main/src/app/state/orderInputSlice.ts#L682

nguvictor:
Fixes #425

Proposed solution is to extract the alphadex-client method and move it into the orderslice

image

### DIFF
--- a/src/app/state/orderInputSlice.ts
+++ b/src/app/state/orderInputSlice.ts
@@ -7,6 +7,7 @@ import {
 import * as adex from "alphadex-sdk-js";
 import { RootState } from "./store";
 import { SdkResult } from "alphadex-sdk-js/lib/models/sdk-result";
+import { getTransactionDetail } from "alphadex-sdk-js/lib/radix-api";
 import { RDT, getRdtOrThrow } from "../subscriptions";
 import { fetchAccountHistory } from "./accountHistorySlice";
 import { fetchBalances } from "./pairSelectorSlice";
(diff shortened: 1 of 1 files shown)


nguvictor:
@fliebenberg do you see any issues with using a customized copy of the alphadex submitTransaction(manifest: any, rdt: RDT) ?

We can use this as an in between solution, until alphadex can support custom messaging in the transaction manifest.

dcts (approved):
Nice! Tested locally and works as expected!
Really cool idea to have the function duplicated until alphadex adds functionality to specify a message!

Another question: should we have "Order Using DeXter" for all orders or maybe we want to differentiate between "Market Order Using DeXter" and "Limit Order Using DeXter"? Not sure though, was just a thought.

Radstakes:
I'm unable to check on mainnet, but this doesn't affect the dapp definition right?  In the above Stokenet screenshot it shows "unknown dapp" but I presume that's just Stokenet.

I think having a single message for both limit and market is fine.

dcts:
> I'm unable to check on mainnet, but this doesn't affect the dapp definition right? In the above Stokenet screenshot it shows "unknown dapp" but I presume that's just Stokenet.
Yes I just tested on mainnet and it shows "DeXter" as expected.

Issue #445, opened 2024-06-04
Remove trade history once user loggs out


dcts:
# To reproduce
1. open dexteronradix.com/trade
2. connect your wallet with an account that has past orders (order history)
3. check that your past orders show inside the "order history" section
4. disconnect your account
5. the order history still shows even though the user is logged out

PR #450, opened 2024-06-11
[Fix]: Remove trade history on log out

restylianos:
Resolves https://github.com/DeXter-on-Radix/website/issues/445

### DIFF
--- a/src/app/state/accountHistorySlice.ts
+++ b/src/app/state/accountHistorySlice.ts
@@ -112,11 +112,11 @@ export const accountHistorySlice = createSlice({
       const adexState = action.payload;
       state.trades = adexState.currentPairTrades;
     },
+    resetAccountHistory: () => {
+      return initialState;
+    },
     setSelectedTable: (state, action: PayloadAction<Tables>) => {
       state.selectedTable = action.payload;
-    },
-    resetOrderHistory: (state) => {
-      state.orderHistory = [];
     },
   },
 
--- a/src/app/state/accountHistorySlice.ts
+++ b/src/app/state/accountHistorySlice.ts
@@ -115,6 +115,9 @@ export const accountHistorySlice = createSlice({
     setSelectedTable: (state, action: PayloadAction<Tables>) => {
       state.selectedTable = action.payload;
     },
+    resetOrderHistory: (state) => {
+      state.orderHistory = [];
+    },
   },
 
   extraReducers: (builder) => {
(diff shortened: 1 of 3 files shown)


src/app/state/accountHistorySlice.ts
```diff
@@ -115,6 +115,9 @@ export const accountHistorySlice = createSlice({
     setSelectedTable: (state, action: PayloadAction<Tables>) => {
       state.selectedTable = action.payload;
     },
+    resetOrderHistory: (state) => {
+      state.orderHistory = [];
```
dcts, line 119: Can you remove the trades as well?
```suggestion
      state.orderHistory = [];
      state.trades = [];
```

dcts:
I noticed that the balance too is not resettet after logout, can you create another reducer method `resetBalances` inside pairSelectorSlice that deletes the balance property on both TokenInfos?
```
    resetBalances: (
      state: PairSelectorState
    ) => {
      delete state.token1.balance;
      delete state.token2.balance;
    },
```

You can also do another PR if you prefer! :)

src/app/state/pairSelectorSlice.ts
```diff
@@ -143,6 +143,9 @@ export const pairSelectorSlice = createSlice({
         state.token2 = { ...state.token2, balance };
       }
     },
+    resetBalance: () => {
+      return initialState;
```
dcts, line 147: Can you only reset the token1 and token2 balances? Otherwise the pair selection will be resettet too, which I don't think is what we want :)

e.g. either delete or set to undefined, but I'd prefere to restore to the default state, which is that the field is not present at all:
```js
resetBalances: (
      state: PairSelectorState
    ) => {
      delete state.token1.balance;
      delete state.token2.balance;
    },
```

dcts (changes requested):

src/app/state/pairSelectorSlice.ts
```diff
@@ -143,6 +143,20 @@ export const pairSelectorSlice = createSlice({
         state.token2 = { ...state.token2, balance };
       }
     },
+    resetBalance: (state: PairSelectorState) => {
```
dcts, line 146: Can you rename to `resetBalances` (pluralized)?

dcts (changes requested):
After the renaming is done, everything else looks good to me! Happy to merge! Great work! 🎉🙏

dcts (approved):

Issue #452, opened 2024-06-12
rewards chest image is misplaced in mobile screens


restylianos:
<img width="371" alt="Screenshot 2024-06-12 at 7 46 21 PM" src="https://github.com/DeXter-on-Radix/website/assets/50713608/c45f7208-1cf5-4352-b87f-9776cc5475e9">

PR #453, opened 2024-06-12
[Fix]: rewards chest misplaced in mobile screens

restylianos:
Minor pr that resolves https://github.com/DeXter-on-Radix/website/issues/452


dcts (approved):
Nice, looks better now!

PR #455, opened 2024-06-12
fix rewards page bug

### LINKED ISSUE #454: Rewards Page Table Overlapped by Footer
image
Just noticed this issue when claiming the latest rewards.  If I zoom out to 67% it's fine, but anything more than that (I normally zoom to 80%), the table is clipped.

dcts:
resolves #454

### DIFF
--- a/src/app/rewards/page.tsx
+++ b/src/app/rewards/page.tsx
@@ -21,7 +21,7 @@ import { DexterToast } from "../components/DexterToaster";
 export default function Rewards() {
   const { showSuccessUi } = useAppSelector((state) => state.rewardSlice);
   return (
-    <div className="bg-[#141414] h-screen flex items-center justify-center">
+    <div className="bg-[#141414] grow flex items-center justify-center">
       <div>
         <HeaderComponent />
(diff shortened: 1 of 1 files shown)


dcts:
Please test whether it worked @Radstakes. As you won't have any rewards loaded on stokenet, you most likely will need to artificially extend the rewards card to see if the fix works. 

You can run this script inside your webdev console, which will duplicate content in the rewards card:

```js
// Select Element to Replace
const elementToCopy = document.querySelectorAll(".px-4.py-4 .flex.flex-col")[0];

// Check if the element exists
if (elementToCopy) {
  // Get the parent element
  let parentElement = elementToCopy.parentElement;

  // Copy the element five times
  for (let i = 0; i < 5; i++) {
    // Clone the element
    let clonedElement = elementToCopy.cloneNode(true);

    // Append the cloned element to the parent
    parentElement.appendChild(clonedElement);
  }
}


```

Radstakes:
Bingo! Thanks for the quick fix 
image

Radstakes (approved):

PR #465, opened 2024-06-25
Batch delete orders

dcts:
This PR implements a feature where the user can cancel multiple orders at once, with a maximum capped at 8 orders. 

Features (to test):
- accountHistory table rows now have a hover state for better UX
- batch deletion: select / deselect orders
- batch deletion: delete X orders
- single deletion should still work, the functionality is shown on hover state but is disabled once the user selects the order for batch deletion.

### Preview
batchDeletion-preview


nguvictor:
photo_2024-06-26_15-52-45
image

I appear to have COMPLETED orders in my open orders tab, but for some reason the json is returning PENDING

image

dcts:
@nguvictor does the same happen when you try to delete such an order using the "Delete Single Order" button (on hover)? If yes I suggest to create a separate issue for this, as it seems to be a weird bug regarding the orders being displayed rather than related to the batch deletion feature.

nguvictor (approved):

PR #477, opened 2024-07-13
Add recent trades

dcts:
Added a tab inside the orderbook panel to show "recent trades" of the pair. 

# Preview

image


fliebenberg:
I have replied on Discord. I think it might be better to import the trades in the same way other adex variables are imported in the subscriptions.ts file.

nguvictor (dismissed):

dcts:
@fliebenberg thanks for the pointers, I added the state to the orderBookSlice and updated via subscriptions. Now the data should not be static anymore but updating in realtime! :)

fliebenberg (approved):

Issue #486, opened 2024-07-19
xUSDC max amount input rounds values up instead of down


fliebenberg:
Issue raised on telegram.
Bug:
When I want to sell max xUSDC from my wallet, the amount is rounded up instead of down. AlphaDex supports 5 decimals for xUSDC, and I have 8 decimals of xUSDC in my wallet. So DeXter tries to sell up to 0.xxxx1 more than is in my wallet

Error from wallet:
InsufficientBalance { 
   requested: 562.70537, 
   actual: 562.705367 
}

fliebenberg:
@dcts I assigned to you as you worked extensively on the new order input. I had a quick look, but could only find code where all amounts are rounded to 8 decimals. So could not find where it is rounded to 5 decimals.

dcts:
# Final Bugreport (to follow bug bounty protocol)

The issue was that xUSDC coin has 6 decimals on [radix dashboard](https://dashboard.radixdlt.com/resource/resource_rdx1t4upr78guuapv5ept7d7ptekk9mqhy605zgms33mcszen8l9fac8vf/summary), but only 5 decimals specified on alphadex.

[...]

PR #488, opened 2024-07-21
Fix x usdc precision bug

dcts:
resolved https://github.com/DeXter-on-Radix/website/issues/486.

The issue seems to be that xUSDC coin has 6 decimals on [radix dashboard](https://dashboard.radixdlt.com/resource/resource_rdx1t4upr78guuapv5ept7d7ptekk9mqhy605zgms33mcszen8l9fac8vf/summary), but only 5 decimals specified on alphadex.

image

## Fix

- When the token specified is xUSDC, it will now always ensure we round down to 5 decimals. Note: this can lead to dust amounts being left on the wallet, but theres nothing we can do to prevent that.
- For that I added a functino toFixedRoundDown including tests.
- I could reproduce the error prior to the fix, then was able to resolve once fix was applied.


fliebenberg:
Although this solves the immediate problem for a specific token, I think we should rather make the fix read the max decimals for a token (we should already have this info in the adex tokens data). That way the problem is solved for any future limited-decimal tokens.

dcts:
I thought about that too, but was not sure whether it would introduce bugs for other pairs, as we dont have test coverage for that and dont know for sure if there will be any side effects.

I would rather wait for another instance of this bug instead of pushing a change that is affecting perfectly fine pairs.

dcts:
@fliebenberg I added the fix for all pairs that have less than 18 decimals, to prevent messing around with 18 decimal coins which do not have any issues currently.

fliebenberg:
The fix you are suggesting will not work as it will result in incorrect decimals for tokens with 8 or less decimals. Alphadex actually rounds all tokens with decimals <= 8 to one less decimal. So if the max decimals on radix is 8, the Alphadex max decimals will be 7. That is why xUSDC is 5 on Alphadex and 6 on Radix. The clearest will be to read the value from the pair info for the token. That is exactly why the info is provided in the pair info. The variables in pairInfo is token1MaxDecimals or token2MaxDecimals. This will also ensure the code works with all future versions of Alphadex. E.g. in a future version of Alphadex it will work with the full decimals of the token, which will mean our code will be wrong again if we do a manual calc.

dcts:
I updated to apply maxdecimals rounding from adex to all tokens as per our discussion :)

fliebenberg (approved):

PR #489, opened 2024-07-22
Fix hydration error landing page

dcts:
We had another instance of hydration error on the landing page caused by the language detection:

image

# This PR
- fixes hydration errors on landing page by adding hydration fix to all landing page components
- adds improved documentation on how to handle hydration errors. I newly recommend only fixing those on demand, as its not easy to anticipate when they occur. With this new guide the action plan is pretty straightforward!

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -181,32 +181,34 @@ To add a new banner, follow these steps:
 
 ## Hydration Error Handling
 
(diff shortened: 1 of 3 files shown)


fliebenberg:
Is this change at all related to issue #497? Not sure if solving #497 will help to solve some of these issues as it moves the language detection into the root component? I am not really a REACT expert so just checking.

dcts:
This change is unrelated @fliebenberg . It solves solely the hydration error on the landing page.

nguvictor (approved):

PR #490, opened 2024-07-23
Multi account support

dcts:
WIP.


fliebenberg:
Question, will this now check all my accounts for rewards or will I have to cycle through each account manually to check for rewards?

dcts:
> Question, will this now check all my accounts for rewards or will I have to cycle through each account manually to check for rewards?

You will have to manually cycle through all accounts.

nguvictor:
Changing accounts, doesn't reset the selected cancelled orders
1. Select less then 8 orders to cancel
2. Switch to a different account
3. It doesn't let you select up to 8 orders on the new account
saa

nguvictor:
Refreshing seems to automatically select the first account, but this can be a different issue to fix.
saaaa

dcts:
> Changing accounts, doesn't reset the selected cancelled orders

Really good catch! Fixed it!

> Refreshing seems to automatically select the first account, but this can be a different issue to fix

I agree, will do in a future PR, tracking here: #516

dcts:
> Refreshing seems to automatically select the first account, but this can be a different issue to fix

I just implemented this too. Now the selection is stored inside cookies, whenever you refresh the page it should still persist. 

Also I improved the speed at which the balance is updated, it was previously a bit slow (3-4 seconds, now its <1s).

fliebenberg (approved):
Great work!

PR #491, opened 2024-07-24
Pairs toggle

### LINKED ISSUE #484: Add toggle to show all trades in Account Hisotry
Currently, the Account Hisotry only shows trades that are related to the currently selected pair. We'd like to introduce a toggle on the top right corner of the AccountHistory component with the label "hide other pairs" which is turned on by default. The user should be able to unselect the toggle, which should then show all trades in all pairs. 

## **Design Mock**

Bildschirmfoto 2024-07-18 um 21 10 42

**Design Figma File**

https://www.figma.com/design/Zjff03BX35fREatqzcWDS7/DeXter-UI?node-id=825-6498&t=gGYTb36pl7VTuLuV-1

saidam90:
Toggle feature "Hide other pairs" to show only selected pairs when enabled and all pairs when disabled.

resolves #484.


src/app/state/accountHistorySlice.ts
```diff
@@ -65,6 +69,51 @@ export const fetchAccountHistory = createAsyncThunk<
   }
 });
 
+export const fetchAccountHistoryAllPairs = createAsyncThunk<
+  SdkResult,
```
dcts, line 73: You can simply return `adex.OrderReceipt[]` here, it will simplify things.
```suggestion
  adex.OrderReceipt[],
```

src/app/state/accountHistorySlice.ts
```diff
@@ -65,6 +69,51 @@ export const fetchAccountHistory = createAsyncThunk<
...
+    return {
+      status: "SUCCESS",
+      data: allOrders,
+    } as SdkResult;
+  } catch (error) {
+    return {
+      status: "ERROR",
+      message: "Failed to fetch account history",
+    } as SdkResult;
+  }
```
dcts, line 114: Once the try/catch block is removed, you can simply return `allOrders`

src/app/state/accountHistorySlice.ts
```diff
@@ -170,15 +219,31 @@ export const accountHistorySlice = createSlice({
...
+    builder
+      .addCase(fetchAccountHistory.fulfilled, (state, action) => {
+        state.orderHistory = action.payload.data.orders;
+      })
+      .addCase(batchCancel.fulfilled, (state) => {
+        state.selectedOrdersToCancel = {};
+      })
+      .addCase(
+        fetchAccountHistoryAllPairs.fulfilled,
+        (state, action: PayloadAction<SdkResult>) => {
```
dcts, line 237:
```suggestion
        (state, action: PayloadAction<adex.OrderReceipt[]>) => {
```

src/app/state/accountHistorySlice.ts
```diff
@@ -170,15 +219,31 @@ export const accountHistorySlice = createSlice({
...
+          if (action.payload.status === "SUCCESS") {
+            state.orderHistoryAllPairs = action.payload.data;
+          } else {
+            console.error(
+              `Failed to fetch account history: ${action.payload.message}`
+            );
+          }
```
dcts, line 244: Inside the `fulfilled` case, only handle success case:
```suggestion
          state.orderHistoryAllPairs = action.payload.data;
```

dcts (changes requested):
Great work! Amazing you pulled this off so fast. 

I added some minor improvements, once fixed I'll review the rest of the code, I think there is a small optimization we can do but will comment later to not have too many open comments at once!

src/app/state/accountHistorySlice.ts
```diff
@@ -31,6 +31,10 @@ export interface AccountHistoryState {
   selectedTable: Tables;
   tables: Tables[];
   selectedOrdersToCancel: Record<string, Order>; // the key is `${orderRecieptAddress}_${nftRecieptId}`
+  hideOtherPairs: boolean;
+  orderHistoryAllPairs: adex.OrderReceipt[];
+  error: string | null;
```
dcts, line 36: You could make the error optional so you don't have to initialize it
```suggestion
  error?: string;
```

src/app/state/accountHistorySlice.ts
```diff
@@ -31,6 +31,10 @@ export interface AccountHistoryState {
...
+  loading: boolean;
```
dcts, line 37: Can you explain on how you intent to use `loading`? Does it represent loading state for `fetchAccountHistory` or `fetchAccountHistoryAllPairs`, or both?

Keep in mind that we call "fetchAccountHistory" every 5 seconds to keep data updated, so if you plan on showing a loading UI using that state, we need to distinguish between calls the user has actively made and automated periodic calls.

src/app/state/accountHistorySlice.ts
```diff
@@ -170,15 +211,30 @@ export const accountHistorySlice = createSlice({
...
+        state.orderHistory = action.payload.data.orders;
+      })
+      .addCase(fetchAccountHistory.pending, (state) => {
+        state.loading = true;
+        state.error = null;
+      })
+      .addCase(fetchAccountHistory.rejected, (state, action) => {
+        state.loading = false;
+        state.error = action.error.message || "Failed to fetch account history";
+      })
```
dcts, line 231: Did you intend to add those cases for `fetchAccountHistoryAllPairs`?

src/app/state/accountHistorySlice.ts
saidam90: I thought that I need to add the cases for `rejected` and `pending` as well, should I remove these? I added `loading` solely for the `pending` case :)

src/app/state/accountHistorySlice.ts
saidam90: Screenshot 2024-07-25 at 21 20 58

I added those cases in response to your comment above, I thought they are for `fetchAccountHistory` 😅 
I think I get it now, so you just sent all 3 cases as a reference, and in our case we only need to catch the error inside the extraReducers - in the `fetchAccountHistoryAllPairs.rejected` (not `fetchAccountHistory.rejected` as I did)

src/app/state/accountHistorySlice.ts
dcts: Oh sorry my bad. No you just need a fulfilled and a rejected case. I would not worry about pending yet, we can add functionality for it if loading takes long but I would remove them for now.

dcts: I suggest to only log the error to the console in the rejected case.

src/app/state/accountHistorySlice.ts
dcts: Yes exactly! Sorry for the confusion!

dcts:
There is a bug, currently when showing all pairs, it will also show all open orders (e.g. orders with status PENDING). We should filter those out.

src/app/components/AccountHistory.tsx
```diff
@@ -198,6 +204,32 @@ function DisplayTable() {
...
+    }
+
+    const intervalId = setInterval(() => {
+      if (showAllPairs) {
+        dispatch(fetchAccountHistoryAllPairs());
+      }
+    }, 120000); // Dispatch every 2 mins (120 seconds)
+
+    return () => clearInterval(intervalId); // Cleanup interval on component unmount
+  }, [dispatch, hideOtherPairs]);
```
dcts, line 226: Can we move this code one layer up to be right after [this code](https://github.com/DeXter-on-Radix/website/blob/main/src/app/trade/page.tsx#L45-L49)?

src/app/components/AccountHistory.tsx
```diff
@@ -198,6 +204,32 @@ function DisplayTable() {
...
+
+  // console.log("combinedOrderHistory:", combinedOrderHistory);
```
dcts, line 228: remove

src/app/state/accountHistorySlice.ts
```diff
@@ -212,4 +259,24 @@ export const selectOrderHistory = createSelector(
...
 );
 
+export const selectCurrentPairAddress = (state: RootState) =>
+  state.pairSelector.address;
+
+export const selectCombinedOrderHistory = createSelector(
+  (state: RootState) => state.accountHistory.orderHistory,
+  (state: RootState) => state.accountHistory.orderHistoryAllPairs,
+  (state: RootState) => state.accountHistory.hideOtherPairs,
+  (state: RootState) => selectCurrentPairAddress(state),
```
dcts, line 269: `selectCurrentPairAddress(state)` seems to be not used, maybe remove if not needed.

src/app/state/accountHistorySlice.ts
```diff
@@ -212,4 +259,24 @@ export const selectOrderHistory = createSelector(
...
+  (orderHistory, orderHistoryAllPairs, hideOtherPairs) => {
+    if (hideOtherPairs) {
+      return orderHistory;
+    } else {
+      return [...orderHistoryAllPairs];
+    }
```
dcts, line 275: refactor as specified in [this comment](https://github.com/DeXter-on-Radix/website/pull/491#issuecomment-2251574960)

src/app/state/accountHistorySlice.ts
```diff
@@ -65,6 +69,39 @@ export const fetchAccountHistory = createAsyncThunk<
...
 
+export const fetchAccountHistoryAllPairs = createAsyncThunk<
+  adex.OrderReceipt[],
+  undefined,
+  { state: RootState }
+>("accountHistory/fetchAccountHistoryAllPairs", async (_, thunkAPI) => {
+  const state = thunkAPI.getState();
+  const pairAddresses = state.pairSelector.pairsList.map(
+    (pairInfo) => pairInfo.address
+  );
```
dcts, line 80: Can you filter out the currently selected pair? something like:
```ts
.filter((pairInfo) => pairInfo.address !== state.pairSelector.address);
```

dcts:
> > 1. There is a bug, currently when showing all pairs, it will also show all open orders (e.g. orders with status PENDING). We should filter those out.
> 
> This could be quite useful. Is it not expected that if I want to look at the history for all pairs I might also want to look at all my open orders? I think we could have the toggle button that applies to both open and historic orders.

@fliebenberg I agree we need to show all orders from all pairs, not only for "Order History" but also for "Open Orders". But the orders still should be filtered correclty, meaning that there should be no PENDING orders inside "Order History" (as its currently the case), but instead show up inside the "Open Orders" tab. 

@saidam90 feel free to start implementing this :)

PR #496, opened 2024-07-25
Add GPLv3 license to project (#493)

### LINKED ISSUE #493: Add GPL license to project
As discussed in #37 , we should add a LICENSE file to the repository using a GPL License.

You can get the license text from an online source, there should be plenty out there.

SiegfriedBz:
This pull request adds the GPLv3 license to the project.

**Note:** I used "Copyright (C) 2024 DeXter-on-Radix" in the LICENSE file. 
Should this be updated to reflect a specific individual or company name instead, such as "Copyright (C) 2024 [Your Real Name or Company Name]"? 
Feedback on this would be appreciated.

resolves #493


fliebenberg:
I think the "Dexter-on-Radix is fine for now. We are a community project rather than a company, so as long as the description is unique to our project it should serve the purpose.

fliebenberg:
Just to understand the license agreement in in laymen's terms: This allows anyone to copy the code, but they need to mention that it was copied from Dexter? Would they be allowed to use it for commercial purposes?

SiegfriedBz:
Yes, as far as I understand, a GPL-licensed software allows anyone to use it for commercial purposes. With restrictions. One being that any software derived from a GPL-licensed software will inherit the same constraints.

Key Points of GPL Licenses
Commercial Use:
Allowed: You can use, modify, and distribute GPL-licensed software for commercial purposes. This means you can sell software that is licensed under the GPL.

Distribution:
If you distribute GPL-licensed software or derivative works, you must comply with the terms of the GPL. This typically involves:
Providing the source code or making it available.
Including a copy of the GPL license with the software.
Ensuring that derivative works are also licensed under the GPL (i.e., the "copyleft" provision).

dcts (commented):
I think we can remove lines 1-5, no?

Otherwise I think this is great. 

Do we need to create a poll in telegram to get community approval for this @fliebenberg ? I personally think this is common sense to have the code being openly used under those conditions, but wanted to hear a second opinion :)

dcts:
> Just to understand the license agreement in in laymen's terms: This allows anyone to copy the code, but they need to mention that it was copied from Dexter

Chat GPTs says this:
Yes, in laymen's terms, this license allows anyone to copy, modify, and distribute the code, but they must credit the original source, DeXter-on-Radix, and include the same license terms with their distribution.

LICENSE
```diff
@@ -0,0 +1,231 @@
+COPYING -- Describes the terms under which DeXter-on-Radix is distributed.
+A copy of the GNU GPL is appended to this file.
+
+IMPORTANT DeXter-on-Radix LICENSE TERMS
+
```
dcts, line 5: Do we really need these 5 lines, IMO we could remove those as they do not add anything that's already covered in the license.

dcts (changes requested):

LICENSE
SiegfriedBz: As you prefer, I added it as a header, because I saw other GPL-licensed projects doing so

dcts: We should remove all lines up to 11 and keep only the official license text. The previous content is redundant and repeats what's in the license (please correct me.if I'm wrong). Other projects often use a LICENSE file for the official text and a separate COPYING file for additional requirements. 

I'm confused why we include COPYING information inside the LICENSE file and why we have it at all without extra requirements. It's clearer to have just the license text so developers know exactly which license applies with no additional conditions.

dcts: That being said, I am open to keep lines 6-11, as they serve as a nice summary of the license itself. So people that do not know about licenses can get a quick overview.

dcts (approved):
Looks good to me! Thanks for tackling this!

PR #497, opened 2024-07-25
Make rootlayout server component

SiegfriedBz:
Refactor RootLayout: Transform from Client to Server Component
- remove use client directive from RootLayout
- extract the store provider and the AppBody components into client components
- update dependency tree


dcts:
closing duplicate of #515

PR #498, opened 2024-07-25
Add Export Orders as CSV Feature

### LINKED ISSUE #495: Export orders as CSV
## Problem

Currently there is no way of exporting orders (open or history) from dexter. 

## Your Task

I suggest to add a small secondary button "export as csv" below all tables, that allow users to export the data inside the table as CSV. 

Bildschirmfoto 2024-07-25 um 00 44 45

## Data

The data of all orders can be retrieved from our redux state, especially from `accountHistory` -> `orderHistory`.

```js
const { orderHistory } = useAppSelector((state) => state.accountHistory);
console.log({ orderHistory });
```

SiegfriedBz:
Resolves #495


src/app/components/AccountHistory.tsx
```diff
@@ -239,6 +240,11 @@ function DisplayTable() {
         </thead>
         <tbody>{tableToShow.rows}</tbody>
       </table>
+      {selectedTable === Tables.ORDER_HISTORY && (
+        <div className="flex m-2">
+          <ExportCsvButton />
+        </div>
+      )}
```
dcts, line 247: Currently the export button is misaligned with the table. To fix this I suggest to create an empy row and add the button as last <td> element of that row. This would mean to remove the element here and refactor your <tbody> as follows:
```jsx
        <tbody>
          {tableToShow.rows}
          {selectedTable === Tables.ORDER_HISTORY && (
            <tr className="!bg-transparent">
              {headers[Tables.ORDER_HISTORY].slice(1).map((_, index) => (
                <td key={index}></td>
              ))}
              <td>
                <ExportCsvButton />
              </td>
            </tr>
          )}
        </tbody>
```

src/app/components/ExportCsvButton.tsx
dcts, line 1: Since the exportCsvButton is dependant on the accountHistory's data, can you move this component inside of "AccountHistory.tsx"? 

I know the file is getting big, but I prefer having all code that belongs together in a single file, and it's also how we usually have the code structured. 

If you want to make the export button an independant component (where the data is being passed in a generalized format as prop), I would support keeping it in a separate file.

dcts (changes requested):
Amazing, works as expected! 

When you can adapt the 2 adressed issues (alignment of button + keep code in the same file) I'll happily merge!

src/app/components/ExportCsvButton.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  // do not display the button if orderHistory is empty
+  if (!orderHistory || orderHistory.length === 0) {
+    return null;
+  }
+
+  return (
+    <button
+      onClick={handleExport}
+      className="text-sm ml-auto hover:text-dexter-gradient-green"
```
dcts, line 53: Can you add "font-bold" class to make the button stand out a bit?

src/app/components/AccountHistory.tsx
SiegfriedBz: Yes that makes perfect sense, thanks ! 
Also, I made it responsive to make the btn always visible without having to x-scroll (left-aligned up to lg screen)

src/app/components/ExportCsvButton.tsx
SiegfriedBz: Yes, ExportCsvButton component now in AccountHistory.tsx

src/app/components/ExportCsvButton.tsx
SiegfriedBz: Done, and I also added whitespace-nowrap to keep the btn content in 1 line

dcts:
closing duplicate of #511

PR #515, opened 2024-07-29
make rootlayout server component

SiegfriedBz:
Refactor RootLayout: Transform from Client to Server Component
- remove use client directive from RootLayout
- extract the store provider and the AppBody components into client components
- update dependency tree


dcts:
@SiegfriedBz can you turn this into a PR draft until we have tested/discussion this? :)

SiegfriedBz:
I made this PR 3mo ago when I started to work on this project.
Best to keep the root layout as a server-component and limit client-components to where client-specific behavior is essential. This approach leverages server-side rendering optimization while keeping your app performant.
Or maybe you should have built this app on Vanilla React and not Next.js ?
Anyway @dcts as you said you are not interested in learning tech but just here to make quick money. 
Good luck with that.

PR #521, opened 2024-08-01
add script to fetch all users

### LINKED ISSUE #517: Fetch all users script
# Problem

Currently, we lack information on the number of unique users and their activity on DeXter. Although all data is publicly available on the ledger, it needs to be aggregated and analyzed. We should also display this data on the landing page in the stats widget (instead of showing an aggregated follower count).

# Your Task

Implement a script using the Alphadex JS API to fetch all orders for all pairs. Store unique addresses in a hashmap and count the number of orders per address.

# Impelementation hints

- IMO its best to create a standalone script inside `scripts/fetchUsers.js`. You can either use JS or TS, whatever you prefer.
- Use the [alphadex JS API](https://socket.dev/npm/package/alphadex-sdk-js). 
- First you will need to get an array of all [PairInfo](https://socket.dev/npm/package/alphadex-sdk-js#pairinfo-classtype) objects.
- The you will need to find out `lastOrderId` and `address` of the pair, which are needed to construct the inputs for [`getPairOrders()`](https://socket.dev/npm/package/alphadex-sdk-js#getpairorders-async-function-back-to-functions-list).
- I am not sure how to get `lastOrderId`, see [discussion on discord](https://discord.com/channels/1125689933735145503/1267244455153504291/1267865097846390880)
- I am not sure what the quota is for the API (how many requests we can trigger per minute), see [discussion on discord](https://discord.com/channels/1125689933735145503/1267244455153504291/1267873267402932225)

### Code snippet (did not test, please modify as needed)

```js
// import all the sdk functions and variables as adex
import * as adex from "alphadex-sdk-js";
...
// run the sdk init function to establish a live connection with the exchange
adex.init();

// use the variables with the live data in your program/app
let adexState = adex.clientState;

// get all PairInfo
const pairsList = adexState.pairsList;

// Exampele for a single pair
const pairInfo = pairsList[0];
const pairAddress = pairInfo.address;
const pairLastOrderId = 10; // TODO: find out how to get this data
const allOrderIds = [...Array(pairLastOrderId).keys()]; // => [1,2,3,...,pariLastOrderId]

// fetch all orders for that pair
const result await adex.getPairOrders(pairAddress, allOrderIds);
console.log(`first 5 orders or pair ${pairAddress}:`);
console.log(result);

// TODO: For each order, find out the wallet address of the creator, and store inside a hashmap/object
const users = {};
// example
users[radixWalletAddress] = users[radixWalletAddress] ? users[radixWalletAddress]+1 : 1;

// TODO: iterate over all pairs

// TODO: ideally the script should show logs (progress bar)

// TODO: I suggest to make the script syncronous at fist
// and move to async later if it takes too long. 
// making it syncronous at the beginning will make sure
// we don't run into API quota issues (hopefully)
```

SiegfriedBz:
Resolves #517 

fetch all users account addresses, returns an object with 2 properties 
{ usersFreqCounter, totalUsersCount }

When I run it in the StatsWidget component,  I get these logs below
best will be to run it server-side

Screenshot 2024-08-01 at 11 07 45

Screenshot 2024-08-01 at 11 07 06


fliebenberg:
Looks good, thanks @SiegfriedBz. 
At the moment we are considering all orders, not just orders placed on Dexter. This is fine for now as 99.9% of AlphaDEX orders are placed through Dexter. But this could change in future if another project also decides to build on AlphaDEX. 
To be accurate, we just need to consider orders with a platformId (one of the order fields) of 4 (Dexter's platform ID). Might as well implement this extra little check now to make sure it is future ready. It wont change the numbers much.

fliebenberg:
I dont know next that well. Is it relatively easy to make this run server side? Also, we probably dont need to update these stats that regularly and could update it by only reading any new orderids since the last update.
I would suggest e update these stats no more than once a day (we could even do it less frequently).

scripts/fetchUsers.js
```diff
@@ -0,0 +1,123 @@
...
+          acc[radixWalletAddress] = (acc[radixWalletAddress] || 0) + 1;
+        }
+      }
+      return acc;
+    }, {});
+
+    // get total unique users count
+    const totalUsersCount = Object.values(usersFreqCounter).reduce(
+      (acc, curr) => {
+        return acc + curr;
```
fliebenberg, line 43: Will this not count the total number of orders, rather than the total number of unique users (accounts)? Or maybe I am understanding it wrong.
To get the number of accounts/users you can just take Array.from(Object.values(usersFreqCounter)).length

SiegfriedBz: Array.from(Object.values(usersFreqCounter)).length
Will this sum up all values for all the keys ? I believe it sums up the number of key/value pairs ? as would do Array.from(Object.keys(usersFreqCounter)).length

usersFreqCounter is an object with keys being unique accounts, and values being the sum of all orders passed by each corresponding unique account
so my totalUsersCount just sum up, for all unique accounts, all orders

dcts: I agree with @fliebenberg , here's a recap:
```js
// To count all users
const totalUsersCount = Object.keys(usersFreqCounter).length;

// To sum all counts of orders placed
const totalOrdersCount = Object.values(usersFreqCounter).reduce((a,b) => a+b);
```

SiegfriedBz: Yes. my big mistake.. I will fix this, and we actually don t need the frequency counter..

scripts/fetchUsers.js
```diff
@@ -0,0 +1,147 @@
...
+  const allOrders = allOrdersByPairs.flat();
+
+  // Process orders to count wallet addresses
+  const usersDict = {};
+  for (let i = 0; i < allOrders.length; i++) {
+    const radixWalletAddress = allOrders[i].settlementAccount; // account address for this order
+    if (!radixWalletAddress) {
+      continue;
+    }
+    usersDict[radixWalletAddress] = usersDict[radixWalletAddress]
```
fliebenberg, line 26: Should the = not be a ==?

PR #527, opened 2024-08-01
fix AnimatedCounter label css

SiegfriedBz:
center AnimatedCounter label in StatsWidget to fix the following issue:

Bildschirmfoto 2024-08-02 um 01 15 32

### DIFF
--- a/src/app/components/StatsWidget.tsx
+++ b/src/app/components/StatsWidget.tsx
@@ -61,7 +61,7 @@ const StatsWidget = () => {
             label={<span className="uppercase">{t("users_trust_us")}</span>}
             wrapperClassName="w-full flex justify-center max-md:flex-col max-md:text-3xl md:text-4xl items-center gap-2 max-md:gap-4"
             counterClassName="text-dexter-green-OG font-bold min-w-44 max-md:text-center md:text-right"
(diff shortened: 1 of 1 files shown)


src/app/components/StatsWidget.tsx
```diff
@@ -61,7 +61,7 @@ const StatsWidget = () => {
             label={<span className="uppercase">{t("users_trust_us")}</span>}
             wrapperClassName="w-full flex justify-center max-md:flex-col max-md:text-3xl md:text-4xl items-center gap-2 max-md:gap-4"
             counterClassName="text-dexter-green-OG font-bold min-w-44 max-md:text-center md:text-right"
-            labelClassName="ml-2"
+            labelClassName="md:ml-2 text-center"
```
dcts, line 64: Can you replace `md:ml-2` with `px-4` please? It solves (1) minimal padding on mobile devices (2) spacing between number and "Users trust use".

dcts (changes requested):

src/app/components/StatsWidget.tsx
SiegfriedBz: Yes, is done

dcts (approved):

PR #531, opened 2024-08-02
Pairs toggle

saidam90:
Added a filter for selectors, sorting by id and fixed the issue with duplicates.


src/app/state/accountHistorySlice.ts
```diff
@@ -212,4 +265,95 @@ export const selectOrderHistory = createSelector(
...
+//         new Date(b.timeSubmitted).getTime() -
+//         new Date(a.timeSubmitted).getTime();
+//       if (timeDifference !== 0) {
+//         return timeDifference;
+//       } else {
+//         return b.id - a.id;
+//       }
+//     });
+//   }
+// );
```
dcts, line 354: Remove outcommented code.

src/app/trade/page.tsx
```diff
@@ -46,11 +53,28 @@ export default function Trade() {
     const intervalId = setInterval(() => {
       dispatch(fetchBalances());
       dispatch(fetchAccountHistory());
+      // console.log("5 seconds");
```
dcts, line 56: remove outcommented code

src/app/trade/page.tsx
```diff
@@ -46,11 +53,28 @@ export default function Trade() {
...
+    const showAllPairs = !hideOtherPairs;
+    if (showAllPairs) {
+      dispatch(fetchAccountHistoryAllPairs());
+    }
+
+    const intervalId = setInterval(() => {
+      if (showAllPairs) {
+        dispatch(fetchAccountHistoryAllPairs());
+      }
+      // console.log("120 seconds");
```
dcts, line 72: remove outcommented code

src/app/trade/page.tsx
```diff
@@ -46,11 +53,28 @@ export default function Trade() {
     const intervalId = setInterval(() => {
       dispatch(fetchBalances());
       dispatch(fetchAccountHistory());
+      // console.log("5 seconds");
     }, 5000); // Dispatch every 5000 milliseconds (5 second)
 
     return () => clearInterval(intervalId); // Cleanup interval on component unmount
   }, [dispatch]);
 
+  useEffect(() => {
```
dcts, line 62: Can you add comments to both useEffect's, for example:

// Update orders of selected pair every 5 seconds
// Update orders of all pairs every 2 mins (if selected)

src/app/state/accountHistorySlice.ts
```diff
@@ -65,6 +69,41 @@ export const fetchAccountHistory = createAsyncThunk<
...
+export const fetchAccountHistoryAllPairs = createAsyncThunk<
+  adex.OrderReceipt[],
+  undefined,
+  { state: RootState }
+>("accountHistory/fetchAccountHistoryAllPairs", async (_, thunkAPI) => {
+  const state = thunkAPI.getState();
+  const pairAddresses = state.pairSelector.pairsList.map(
+    (pairInfo) => pairInfo.address
+  );
+  // .filter((pairAddress) => pairAddress !== state.pairSelector.address);
```
dcts, line 81: remove outcommented code

src/app/state/accountHistorySlice.ts
```diff
@@ -212,4 +265,95 @@ export const selectOrderHistory = createSelector(
...
+
+      // Add orders from orderHistory
+      orderHistory.forEach((order) => {
+        orderMap.set(order.id, order);
+      });
+
+      // Add orders from orderHistoryAllPairs, will overwrite any duplicates from orderHistory
+      orderHistoryAllPairs.forEach((order) => {
+        orderMap.set(order.id, order);
+      });
```
dcts, line 287: Since the order Id's are not unique among pairs but only within a given pair (e.g. every pair starts the ordercount at 1, so every pair has a single order with the id 1), we need to add the name or address to the key of the hashmap, to make the orders truly unique. For example:

```ts
      // Generate unique orderId helper
      const getUniqueOrderId = (order: Order) => `${order.name}_${order.id}`;

      // Add orders from orderHistory
      orderHistory.forEach((order) => {
        orderMap.set(getUniqueOrderId(order), order);
      });

      // Add orders from orderHistoryAllPairs, will overwrite any duplicates from orderHistory
      orderHistoryAllPairs.forEach((order) => {
        orderMap.set(getUniqueOrderId(order), order);
      });
```

dcts (changes requested):
Great work, code is super clean and readable! Also the fetching works perfectly now and all orders appear!

Added some minor comments, please fix.

Also I experience a weird bug, where the ordering in the account history is not correctly working. I will need to run locally and debug with my dev / test account, which has a lot of oders. Will do tomorrow!

(See below the ordering is not correct)
Bildschirmfoto vom 2024-08-04 02-35-39

PR #535, opened 2024-08-03
animate hamburger button and mobile menu on open/close

SiegfriedBz:
add animation on the hamburger menu & animation when mobile menu open/close

### DIFF
--- a/src/app/components/NavBar.tsx
+++ b/src/app/components/NavBar.tsx
@@ -1,5 +1,3 @@
-"use client";
-
 import React, { useEffect, useState, useRef } from "react";
 import Image from "next/image";
 import Link from "next/link";
--- a/src/app/components/NavBar.tsx
+++ b/src/app/components/NavBar.tsx
@@ -1,9 +1,11 @@
+"use client";
+
 import React, { useEffect, useState, useRef } from "react";
 import Image from "next/image";
 import Link from "next/link";
 
 import { useSelector } from "react-redux";
-import { useAppDispatch, useAppSelector } from "hooks";
+import { useAppDispatch, useAppSelector, useHydrationErrorFix } from "hooks";
 import { getSupportedLanguagesAsString } from "../state/i18nSlice";
 
 import { i18nSlice } from "../state/i18nSlice";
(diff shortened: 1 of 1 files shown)


dcts:
Nice, I really like the fade-in (zoom-in) effect when the menu opens, good addition making the UX more smooth! Great improvement!

There are some things I noticed that should be addressed:

1. The padding is more narrow when vhecking on mobile, see screenshots: 
IMG_20240803_131819
IMG_20240803_131922

2. The animation of the hamburger menu is great too, but it looks off, as it seems those are 2 instances of the animation that blend in. Is that the case?

SiegfriedBz:
@dcts you have tiger eyes ;) Yes there are 2 instances of the btn. I m going to see how I can make a single one, and if I don t find out how to, ok I will remove the animation on the btn
Ok to increase the padding on the btn
I keep the animation on the mobile menu

SiegfriedBz:
I removed the duplicated btn, it looks better
but I just saw that depending on where we click (on the btn), it can take 2-3 attempts to close the menu, I m gonna try to fix this.
Please give me your feedback.

dcts:
@SiegfriedBz yes the transition now looks perfect (tested on mobile)!! 💯🔥🎉
But yes I also noticed it sometimes does not close as expected. Also the padding is still very narrow:
IMG_20240803_165333

dcts:
@SiegfriedBz When testing this on the deploy preview I am running into a pretty sever bug, where the website fully crashes with no console errors shown and simply shows the error code 11. 

See video (in the video I am reloading the page multiple time):

[Bildschirmaufzeichnung vom 04.08.2024, 14:22:48.webm](https://github.com/user-attachments/assets/250776c2-2b3d-4645-9cd5-d3a3456c3dd6)

Does anyone else run into this issue? I tested it across browsers and with both deploy preview URLs and it happens all the time. It even happens when I checkout this branch locally and run it on my device.

src/app/components/NavBar.tsx
```diff
@@ -1,9 +1,11 @@
+"use client";
+
```
dcts, line 2: Can you remove this, not sure if this is causing the site to crash?

dcts: Tested locally on my device, even when removed, the site crashes.

SiegfriedBz: I just removed  the "use client" directive but it s very strange that the error would come from this..

On my end when I visit both both deploy preview URLs below it s working fine
Screenshot 2024-08-04 at 20 38 02

yuliaSharabi:
When you switch orintation, you will not see teh hamburger anymore, just the regular connect button. 
See the video I've taken.

https://github.com/user-attachments/assets/878d0957-1ae3-4470-b0c2-05e82281eae3

SiegfriedBz:
@yuliaSharabi this is intended, and not orientation dependent, but screen width dependent. We want to display the burger btn only on mobile

yuliaSharabi:
@SiegfriedBz The video record is from my mobile phone, therefore IT IS ON MOBILE.
In this case when the screen is wider on mobile, should it not show the button as Radix icon, if not hamnurger. It shouldn't display it as "connect" I belive?

dcts:
> @SiegfriedBz The video record is from my mobile phone, therefore IT IS ON MOBILE. In this case when the screen is wider on mobile, should it not show the button as Radix icon, if not hamnurger. It shouldn't display it as "connect" I belive?

@yuliaSharabi the design works as expected. Responsiveness is designed based on screenwidth, meaning if there is enough screen real estate to show the "connect" button we will show it.

dcts (approved):

Issue #432, opened 2024-05-30
Show resource addresses


dcts:
Currently there is no way for the user to determine the resource address of the traded tokens. When we allow permissionless listing on dexter as its currently the case we run into the potential scenario that a duplicate fake coin could be created and the user would have no way to verify the legitimacy of the coin by checking the resource address. 

We should add the resource address in the UI. I suggest to create an "Info" tab next to "PriceChart".

dcts:
I suggest to show the following data inside the INFO tab:
- token 1 + token 2:
  - icon url
  - ressourcce address (including link to radix-dashboard
  - name
  - symbol
- information on the adex pair:
  - address: The on-ledger address of the pair component for the pair
  - rderReceiptAddress: The on-ledger resource address of the order receipt NFT for the pair
  - name: Name for the pair. Will usually be a combination of the symbols of the tokens in the pair. e.g. ADEX/XRD

dcts:
The Design for this feature is now finished!

![image](https://github.com/user-attachments/assets/a1997641-912f-4522-852f-8cafec4fd390)

Figma File:
https://www.figma.com/design/Zjff03BX35fREatqzcWDS7/DeXter-UI?node-id=0-1&t=zvlr18AoVEWZQfqt-1

The file is full with designs, so its hard to find, but here a roadmap: 
(1) go to "Prototypes and tests"
(2) right unter voting app design you can find this design (see below). 
![image](https://github.com/user-attachments/assets/8659e8e8-b788-4cde-a855-4d210a4d6c8d)


# How to get the data from redux

The data for this is already available, you can get it by combining data from the rewardSlice and PairSelectoSlice. You can follow this logic:
```ts
// 1. get array of all pairs
const { pairsList } = useAppSelector((state) => state.rewardSlice);

// 2. get currenlty selected pair address
const selectedPairAddress = useAppSelector((state) => state.pairSelector.address);

// 3. get the relevant pairInfo
const pairInfo = pairsList.find((pairInfo) => pairInfo.address === selectedPairAddress);
const { name, address, orderReceiptAddress, token1, token2,  } = pairInfo;
```

fliebenberg:
This would be a great addition. I would suggest we call the tab "Pair Info" rather than "Token Info" as indicated on the design.

saidam90:
You can assign this to me 🤚🏼

PR #544, opened 2024-08-06
Add PairInfo tab

saidam90:
solves #432 

Added a tab for PairInfo.


dcts (changes requested):
Looks great overall! 

Some improvement proposals:
1. Can you hide the candle timeselection whenever the pair info is selected?
Bildschirmfoto 2024-08-06 um 18 46 33

2. Is it possible to always keep the tabs next to each other, and instead wrap the candle timeframe selection?
Bildschirmfoto 2024-08-06 um 18 52 03

dcts (changes requested):
Some more improvements:

3. The two coin boxes should both take up exactly 50% of the full space, to align with the orderrInput and orderBook components. Currently left box it too small for viewport width of 740px, and for 840px viewportwidth the right box is too small. I think you can just add a "flex-grow" on both elements, but probably will need to fix also oversizing in case the content is too big. 

Bildschirmfoto 2024-08-07 um 00 32 31

4. I think to be consistent, all those elements marked red should have the same bottom margin. I suggest `mb-1` for all those elements (maybe you'll need to use `!mb-1` for `<p>` elements. 

Bildschirmfoto 2024-08-07 um 00 42 44

5. Please use the [shortenString function](https://github.com/DeXter-on-Radix/website/blob/29a69e7179492e98bfbab1185119296ffb7f8fea/src/app/utils.ts#L182-L196) to display a shorter version of the addresses, I suggest to use 20 for `showEnd` and to set `showStart` in a way that will show "resource..." or "component...", so either 8 or 9. This is a pattern shared among all crypto space, as people usually check the last few characters of an address.

6. Can you add `px-2` to both nav elements (Price chart + Pair info) to space out the tabs a bit?
Bildschirmfoto 2024-08-07 um 00 51 34


7. Not neccesarily needed, but when copying it would be awesome if the user would get a feedback, something similar to this (found [here](https://dashboard.radixdlt.com/resource/resource_rdx1tknxxxxxxxxxradxrdxxxxxxxxx009923554798xxxxxxxxxradxrd/summary))
Bildschirmfoto 2024-08-07 um 00 45 19

src/app/components/PriceChart.tsx
```diff
@@ -260,12 +262,75 @@ function PriceChartCanvas(props: PriceChartProps) {
   );
 }
 
+enum ChartTabOptions {
+  TRADING_CHART = "TRADING_CHART",
+  PAIR_INFO = "PAIR_INFO",
+}
+
 export function PriceChart() {
```
dcts, line 270: Can you rename this component to `TradingChartOrPairInfo` for clarity? Otherwise its a bit confusing to me what the difference between priceChart and TradingChart is. 

You will also need to adapt the naming inside the trade page where this component is imported and rendered.

src/app/components/PriceChart.tsx
```diff
@@ -316,3 +354,126 @@ export function PriceChart() {
     </>
   );
 }
+
+export function PairInfoTab() {
+  const t = useTranslations();
+  const { pairsList } = useAppSelector((state) => state.rewardSlice);
```
dcts, line 360: We have the `pairsList` available in both the RewardSlice and also pairSelector. I suggest to use the one from pairSelector instead:
```suggestion
  const { pairsList } = useAppSelector((state) => state.pairSelector);
```

src/app/components/PriceChart.tsx
```diff
@@ -316,3 +354,126 @@ export function PriceChart() {
...
+  const pairInfo = pairsList.find(
+    (pairInfo) => pairInfo.address === selectedPairAddress
+  );
+
+  let name, address, orderReceiptAddress, token1, token2;
+  if (pairInfo) {
+    ({ name, address, orderReceiptAddress, token1, token2 } = pairInfo);
+  } else {
+    return "Selected pair not found in pairsList";
+  }
```
dcts, line 373: This is a more concise syntax:
```suggestion
  if (!pairInfo) {
    return "Selected pair not found in pairsList";
  }
  const { name, address, orderReceiptAddress, token1, token2 } = pairInfo;
```

src/app/components/PriceChart.tsx
```diff
@@ -316,3 +354,126 @@ export function PriceChart() {
...
+
+  const shortenAddress = (address: any) => {
+    if (!address) return "";
+    return `${address.slice(0, 40)}...`;
+  };
```
dcts, line 378: Please use [this function](https://github.com/DeXter-on-Radix/website/blob/29a69e7179492e98bfbab1185119296ffb7f8fea/src/app/utils.ts#L182-L196) instead :)

src/app/components/PriceChart.tsx
```diff
@@ -316,3 +354,126 @@ export function PriceChart() {
...
+
+  return (
```
dcts, line 380: This full component could be split up into 3 sub-components for better readibility and maintainability. For example:
- AddressAndLabel
- CoinInfo
- CopyIcon

I shared more details via discord DM.

src/app/components/PriceChart.tsx
```diff
@@ -316,3 +354,126 @@ export function PriceChart() {
...
+            </div>
+          </div>
+
+          <div className="border-b-2 border-[rgba(255,255,255,0.05)] sm:border-r-2 sm:border-b-0 my-4 sm:my-0 sm:mx-4 sm:min-h-[150px] sm:max-h-[200px]"></div>
+
+          <div className="flex flex-col items-start xs:pl-0 sm:pl-4">
+            <div className="flex items-center sm:mb-2 sm:pt-8 xs:pt-4">
+              <img
+                src={token2?.iconUrl}
+                alt=""
```
dcts, line 451: Is there a reason why alt text is shown for token1 but not for token2?

src/app/components/PriceChart.tsx
```diff
@@ -260,12 +262,75 @@ function PriceChartCanvas(props: PriceChartProps) {
...
+        <div className="flex space-x-4 sm:space-x-5 pb-0 pt-2">
+          {[
+            [t("trading_chart"), ChartTabOptions.TRADING_CHART],
+            [t("pair_info"), ChartTabOptions.PAIR_INFO],
+          ].map(([title, tab], indx) => {
+            const isActive = tab === currentTab;
+            return (
+              <span
+                key={indx}
+                className={`text-sm sm:text-base pb-2 sm:pb-3 px-2 ${
```
dcts, line 294: Keep the `text-base` class on mobile for consistency, as the "Order book" and "Recent trades" tabs also maintain the same font size on mobile.
Suggested edit:
```suggestion
                className={`text-base pb-2 sm:pb-3 px-2 ${
```

src/app/components/PriceChart.tsx
```diff
@@ -316,3 +354,116 @@ export function PriceChart() {
...
+  return (
+    <div className="flex flex-col items-start xs:mb-4 mb-4 sm:mb-0 w-[50%]">
+      <div className="flex items-center mb-3 pt-8">
+        <img
+          src={token.iconUrl}
+          alt={token.symbol}
+          className="w-8 h-8 rounded-full"
+        />
+        <p className="pl-2 text-base">
+          {token.name.split(" ")[0]} ({token.symbol})
```
dcts, line 455: Use the full token name. On testnet, it may appear as "Dexter Token", but on mainnet, it is correctly "DeXter".

 ```suggestion
          {token.name} ({token.symbol})
```

**Review by dcts — CHANGES_REQUESTED** [review]
Amazing! Great work with the refactor and all the changes made!

File: package.json [hunk 9]
```diff
@@ -23,6 +23,7 @@
     "@radixdlt/radix-dapp-toolkit": "^2.0.0",
     "@reduxjs/toolkit": "^2.0.1",
     "@tailwindcss/container-queries": "^0.1.1",
+    "@tippyjs/react": "^4.2.6",
```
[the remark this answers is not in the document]
saidam90: noted, can do it once I finish this

src/app/components/PriceChart.tsx
```diff
@@ -260,12 +263,75 @@ function PriceChartCanvas(props: PriceChartProps) {
   );
 }
 
-export function PriceChart() {
+enum ChartTabOptions {
```
dcts, line 266:
```suggestion
enum ChartOrInfoTabOptions {
```

src/app/components/TextToCopy.tsx
```diff
@@ -0,0 +1,46 @@
...
+import Tippy from "@tippyjs/react";
+import "tippy.js/dist/tippy.css";
+import "tippy.js/dist/svg-arrow.css";
+import { useTranslations } from "../hooks";
+
+interface TextToCopyProps {
+  targetUrl: string;
+}
+
+export function TextToCopy({ targetUrl }: TextToCopyProps) {
```
dcts, line 13: Sorry I think this was a misunderstanding, I meat to rename `targetUrl` -> `textToCopy`. The component being named "CopyIcon" was fine.

Please change to:

```suggestion
interface CopyIconProps {
  textToCopy: string;
}

export function CopyIcon({ textToCopy }: CopyIconProps) {
```

dcts (changes requested):

dcts (approved):
🎉

PR #568, opened 2024-08-25
Buy sell panel slider

saidam90:
Implemented Buy/Sell panel percentage slider.


package.json
```diff
@@ -47,6 +47,7 @@
     "resize-observer-polyfill": "^1.5.1",
     "tailwind-merge": "^2.4.0",
     "tailwindcss": "3.3.2",
+    "tippy.js": "^6.3.7",
```
dcts, line 50: Why don't we use "@tippyjs/react" that we already have?

dcts:
Some feedback:
- I tested the following user journeys: 
  - MARKET BUY✅
  - MARKET SELL🔴 (see next bulletpoint)
  - LIMIT BUY✅
  - LIMIT SELL✅
- I cannot use MARKET SELL order anymore, the SELL button seems to be always disabled
Bildschirmfoto 2024-09-08 um 23 28 49

- When using a LIMIT SELL order, I set the slider to 100%, but the tooltip shows 100.00003% (so more than 100%), and the balance then is marked as "insufficient".
Bildschirmfoto 2024-09-08 um 23 32 31

src/app/styles/globals.css
dcts, line 1: Is it possible to make those classes inline tailwind CSS classes instead of using global classes? 
If you run into the issue that they are not applied or overwritten, you can always try adding `!` before the class, so it becomes `!important`.

src/app/components/OrderInput.tsx
```diff
@@ -451,13 +454,60 @@ function SubmitButton() {
...
+    const isXRDToken = isBuyOrder
+      ? token2.symbol === "XRD"
+      : token1.symbol === "XRD";
+    let balance = isBuyOrder ? balanceToken2 : balanceToken1;
+
+    if (newPercentage === 100 && isXRDToken) {
+      balance = Math.max(balance - XRD_FEE_ALLOWANCE, 0);
+    }
+
+    const amount = (balance * newPercentage) / 100;
```
dcts, line 494: Please use our Calculator class for any calculations. The reason is that javascript cannot precisely calculate (you can see that by typing in 0.1+0.2 in a console, the result is surprising. 

Any operation (addition, substraction, division, multiplication) should be done using our [`Calculator`](https://github.com/DeXter-on-Radix/website/blob/main/src/app/services/Calculator.ts) class.

```suggestion
    const amount = Calculator.divide(Calculator.multiply(balance, newPercentage), 100);
```

src/app/components/OrderInput.tsx
```diff
@@ -805,10 +872,163 @@ function InputTooltip({ message }: { message: string }) {
...
 
 // TODO(dcts): implement percentage slider in future PR
-function PercentageSlider() {
-  return <></>;
+interface PercentageSliderProps {
+  initialPercentage: number;
+  callbackOnPercentageUpdate: (newPercentage: number) => void;
+  isLimitOrder: boolean;
+  isBuyOrder: boolean;
+  isSellOrder: boolean;
```
dcts, line 880: Should this be
isMarketBuyOrder
isMarketSellOrder
?

Or are we using those booleans also for LIMIT orders?

src/app/components/OrderInput.tsx
```diff
@@ -805,10 +872,163 @@ function InputTooltip({ message }: { message: string }) {
...
+    sliderRef.current.value = "0";
+    sliderRef.current.style.backgroundSize = `0% 100%`;
+    setPercentage(0);
+
+    if (inputToken1 > 0 && isLimitOrder && isBuyOrder) {
+      return;
+    } else if (inputToken2 > 0 && isLimitOrder && isSellOrder) {
+      return;
+    } else if (balanceToken2 && inputToken2 > 0) {
+      const newPercentage = (inputToken2 / balanceToken2) * 100;
```
dcts, line 925: Use [Calculator](https://github.com/DeXter-on-Radix/website/blob/main/src/app/services/Calculator.ts) class here

src/app/components/OrderInput.tsx
```diff
@@ -805,10 +872,163 @@ function InputTooltip({ message }: { message: string }) {
...
+          <div className="flex justify-between items-center">
+            {Array(5)
+              .fill(0)
+              .map((_, index) => (
+                <span
+                  key={index}
+                  className="dot"
+                  style={
+                    {
+                      left: `${(index * 100) / 5}%`,
```
dcts, line 990: Use [Calculator](https://github.com/DeXter-on-Radix/website/blob/main/src/app/services/Calculator.ts) class here

src/app/components/OrderInput.tsx
```diff
@@ -805,10 +872,163 @@ function InputTooltip({ message }: { message: string }) {
...
+              ))}
+          </div>
+        </div>
+        <div className="w-full">
+          <div className="slider-labels">
+            <div className="flex justify-between text-xxs mt-1 mb-5">
+              <span className="absolute" style={{ left: "0%" }}>
+                0%
+              </span>
+              <span className="absolute" style={{ left: "22%" }}>
```
dcts, line 1003: Is this 22% on purpose (and not 25%)?

src/app/components/OrderInput.tsx
```diff
@@ -805,10 +872,163 @@ function InputTooltip({ message }: { message: string }) {
...
+              </span>
+              <span
+                className="absolute"
+                style={{ left: "50%", transform: "translateX(-50%)" }}
+              >
+                50%
+              </span>
+              <span
+                className="absolute"
+                style={{ left: "74%", transform: "translateX(-50%)" }}
```
dcts, line 1014: Same here, I assume 74% is on purpose (and not 75%)?

dcts (changes requested):

src/app/styles/globals.css
dcts: @saidam90 did you try this?

src/app/components/OrderInput.tsx
dcts: @saidam90 can you comment on this?

src/app/components/OrderInput.tsx
```diff
@@ -451,13 +454,64 @@ function SubmitButton() {
...
   const isBuyOrder = side === "BUY";
   const isSellOrder = side === "SELL";
 
+  useEffect(() => {
+    if (isMarketOrder) {
+      sliderCallback(0, "MARKET");
+    } else if (isLimitOrder) {
+      sliderCallback(0, "LIMIT");
+    }
+  }, [isBuyOrder, isSellOrder, isMarketOrder, isLimitOrder]);
```
dcts, line 481: React Hook useEffect has a missing dependency: 'sliderCallback'. Either include it or remove the dependency array

src/app/components/OrderInput.tsx
```diff
@@ -451,13 +454,64 @@ function SubmitButton() {
...
+
+  const sliderCallback = (newPercentage: number, type: string) => {
+    console.log(`Type is: ${type}`);
```
dcts, line 484: remove console log.

src/app/components/OrderInput.tsx
```diff
@@ -479,7 +541,15 @@ function UserInputContainer() {
         <>
           <CurrencyInputGroup userAction={UserAction.UPDATE_PRICE} />
           <CurrencyInputGroup userAction={UserAction.SET_TOKEN_1} />
-          <PercentageSlider />
```
dcts, line 482: The percentage slider in market orders should be **below** the input field, not above.
Bildschirmfoto 2024-09-19 um 04 35 02

src/app/components/OrderInput.tsx
```diff
@@ -805,10 +876,184 @@ function InputTooltip({ message }: { message: string }) {
...
+    inputToken2,
+    balanceToken2,
+    isLimitOrder,
+    isBuyOrder,
+    isSellOrder,
+  ]);
+
+  return (
+    <>
+      <div className="slider-container rounded-md w-full relative mt-5">
```
dcts, line 955: can you make the full slider opacity of 0.7, currently there is too much emphasis on it.

```suggestion
      <div className="slider-container rounded-md w-full relative mt-5 opacity-70">
```

dcts (changes requested):
Looks great! Using the calculator seems to have removed the issues with rounding I ran into. 

Also the reported bugs seem fixed for me too.

Some minor changes. I'd say let's test this on stokenet with 3-4 testers.

dcts (approved):

Radstakes (approved):

PR #574, opened 2024-09-01
Intuitive pair search

saidam90:
Made search more intuitive.

### DIFF
--- a/src/app/utils.test.ts
+++ b/src/app/utils.test.ts
@@ -6,6 +6,8 @@ import {
   shortenWalletAddress,
 } from "./utils";
 
+import { searchPairs } from "./utils";
+
 // the separators are set to "." and " " for testing purposes
 // inside jest.setup.js
 describe("displayAmount", () => {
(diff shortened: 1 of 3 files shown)


src/app/utils.ts
```diff
@@ -540,3 +540,66 @@ export function getLocalStoragePaginationValue(id?: string) {
 
   return undefined;
 }
+
+export const searchPairs = (query: string, pairsList: any): any => {
```
dcts, line 544: Can you add a comment here, e.g.:

```
// TODO: Update input and return types to `PairInfo[]`. Currently using `any` 
// due to unresolved issues. Investigate the cause of the problem.
```

src/app/utils.test.ts
```diff
@@ -344,3 +346,148 @@ describe("shortenWalletAddress", () => {
...
+      },
+      token2: {
+        name: "Radix",
+        symbol: "XRD",
+      },
+    },
+  ];
+
+  test("should find pairs by full name", () => {
+    const result = searchPairs("DEXTR/XRD", pairsList);
```
dcts, line 388: This means full name of the project, not symbol. So for example XRD is the token of the project Radix. All pairs containing "XRD" should be found when typing "radix" 
```suggestion
    const result = searchPairs("Radix", pairsList);
```

You will need to adapt the expected result as well, as we expect all pairs to be returned (if you use "Radix").

src/app/utils.test.ts
```diff
@@ -344,3 +346,148 @@ describe("shortenWalletAddress", () => {
...
+        },
+        token2: {
+          name: "Radix",
+          symbol: "XRD",
+        },
+      },
+    ]);
+  });
+
+  test("should find pairs by symbol and full name with different case", () => {
```
dcts, line 421: Please add a test for "name with different case" -> e.g. "RaDIx"

src/app/utils.test.ts
```diff
@@ -344,3 +346,148 @@ describe("shortenWalletAddress", () => {
...
+        },
+        token2: {
+          name: "Radix",
+          symbol: "XRD",
+        },
+      },
+    ]);
+  });
+
+  test("should find pairs with swapped order of tokens", () => {
```
dcts, line 455: Can you add another test here to test whether this also works for names? 

e.g. -> "Radix dexter" -> should find "DEXTR/XRD" pair.

src/app/utils.ts
```diff
@@ -540,3 +540,66 @@ export function getLocalStoragePaginationValue(id?: string) {
...
+  const searchQuery = query.trim().toLowerCase().replace(/\s+/g, " ");
+
+  const hasTypoTolerance = (source: string, target: string): boolean => {
```
dcts, line 547: The sliding window approach to check for typos is straight forward and takes into account character mismatches, so its a great start and we can keep it. But for future reference, can we add a TODO here to switch to use the Levenstein distance, which also takes into account missing characters (deletion) or extra characters (insertion), and is a more common approach to handle typo tolerance. 

```ts
// TODO: Implement typo tolerance using Levenshtein distance instead of sliding window approach.
```

src/app/utils.ts
```diff
@@ -540,3 +540,86 @@ export function getLocalStoragePaginationValue(id?: string) {
 
   return undefined;
 }
+
+// TODO: Update input and return types to `PairInfo[]`. Currently using `any` // due to unresolved issues. Investigate the cause of the problem.
```
dcts, line 544:
```suggestion
// TODO: Update input and return types to `PairInfo[]`. Currently using `any`
// due to unresolved issues. Investigate the cause of the problem.
```

dcts (dismissed):
Great work, works perfectly! 
+ kudos for implementing levenstein distance that quickly! 🔥💯

dcts (approved):

Issue #283, opened 2024-03-08
Console errors on pageload


dcts:
We should fix those console errors, they happen on every pageload:

![Bildschirmfoto vom 2024-03-08 16-28-13](https://github.com/DeXter-on-Radix/website/assets/44790691/25c3195d-1750-46ff-89ec-6323e56a133d)

dcts:
To me it looks like this is already resolved even without #577. At least I don't see any errors in the console when checking the live app at dexteronradix.com. 

@SiegfriedBz please re-open if you still can reproduce these errors.

SiegfriedBz:
@dcts 
actually I decided to implement the error.tsx file when I saw the  #283 issue
but I couldn t reproduce specifically the  #283 issue
I believe an error.tsx file is best practice to handle unexpected runtime errors and display a fallback UI, independently of the  #283 issue depicted here

PR #577, opened 2024-09-10
add error boundary

SiegfriedBz:
handle uncaught exceptions and solves #283 

can be tested by throwing an error in any page
 `throw new Error("This is a simulated error!");`

[Untitled_ Sep 10,2024 2_24 PM.webm](https://github.com/user-attachments/assets/7216049e-0082-409b-b850-e81205f66873)

Next.js doc [here](https://nextjs.org/docs/app/building-your-application/routing/error-handling#uncaught-exceptions)


dcts:
Some questions:
1. When I throw inside the console as described above `throw new Error("This is a simulated error!")` nothing happens. Is this expected?
2. Assuming an error gets thrown inside the /trade page, would the UI simply be redirected to "Something went wrong"? 
3. What will happen if you click on "Try again"? Will you be redirected to the page you were on before, or will the action be re-launched somehow?

SiegfriedBz:
@dcts

When using Next.js (also with React alone) I believe it is best practice to implement an error file to handle unexpected runtime errors and display a fallback UI. 
The error.tsx file here (in Next.js) is actually making use of [React Error Boundary](https://react.dev/reference/react/Component#catching-rendering-errors-with-an-error-boundary)

1. As the error.tsx is defined as a client-component, to see the console.log _This is a simulated error!_ you need to look in the browser dev tools

2. As far as I understand the mechanism, this is not a _redirect_ 
If an error is thrown (manually or not) and uncaught in a page, we can see that the url does not change. 
But from this page where this error is thrown and uncaught, this exception "travels up" the component tree until it finds an "error boundary" (error.tsx) and then the error.tsx component is rendered and finally painted to the screen. 

3. Clicking on "Try again" will make React trying to re-render the current page (no redirect involved). If the error is thrown again, the same error UI will be painted, otherwise the expected UI for this page will be rendered and painted.
The purpose of this error.tsx file is to create a better UX in case of an unanticipated / uncaught exception would be thrown, while still being able to log the error for devs

Please have a look at
https://nextjs.org/docs/app/api-reference/file-conventions/error
https://nextjs.org/docs/app/building-your-application/routing/error-handling#uncaught-exceptions

fliebenberg:
Where are we with this PR? Not sure if we are waiting on something or just need to review and merge?

dcts:
> Where are we with this PR? Not sure if we are waiting on something or just need to review and merge?

For this PR to merge I suggest to:
1. improve the styling of the "Something went wrong" text, it looks way too big.
2. If we have access to the error, we should display it to the user so that they can copy the specifics, which will lead to better bugreports.
3. we need to test the user experience when uncaught errors happen, and test the "try again" functionality to ensure everything works properly. For that I suggest to mock potential scenarios where this could happen.

dcts:
Closing due to inactivity. Anyone free to re-open or create a new PR if needed.

PR #588, opened 2024-10-01
Fix division by zero error - Issue_587

yuliaSharabi:
when trade.token1Amount being zero, division by zero occurs on Recent Orders view. This change check whether it's zero, it's return 0.


dcts (approved):
Great work, this fixed the issue!

PR #599, opened 2024-11-02
fix selection not reset when switching pairs

### LINKED ISSUE #592: Cancel orders functionality not working correctly
When I select an order to cancel in one pair, it shows "Cancel 1 Order" in red at the top of the column.
image
When I then navigate to another pair, the red "Cancel 1 Order" is still displayed although I am now in a different pair with different open orders displayed - none of which are selected. If I now select an order here as well it will show "Cancel 2 orders" even though I am only seeing one order selected for the pair.
If I now navigate back to pair 1, it still shows "Cancel 2 orders" but it now shows no orders selected, not even the one I had selected previously. When I select that order again, it shows as selected but does not increase the count again.

I think we should make it that selected orders are reset every time the user changes pairs. So batch cancel should only work for one pair. In fact it will only work for one pair as cancel tx needs to specify a specific pair component.

dcts:
resolves #592

I could replicate and fix locally for me, can you verify it works for you too @fliebenberg ?

### DIFF
--- a/src/app/components/PairSelector.tsx
+++ b/src/app/components/PairSelector.tsx
@@ -7,6 +7,7 @@ import React from "react";
 import { searchPairs } from "utils";
 
 import { BLACKLISTED_PAIRS } from "../data/BLACKLISTED_PAIRS";
+import { accountHistorySlice } from "state/accountHistorySlice";
 
 interface PairInfo {
   name: string;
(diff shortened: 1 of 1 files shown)


fliebenberg (approved):
Yep, working for me now.

PR #620, opened 2024-12-13
FEAT: Automate creating redirects

joysamaddar:


dcts:
Hey! Thanks for the PR. 

Unfortunatley this won't work with our current setup. DeXter is setup as a static webapp using newxt export, which creates a packages build file that is hosted on a CDN. To use redirects you need to enable server capabilities, which we currently don't want to do, in order to reduce costs and improve decentralization. 

The build may work locally, but when I try to deploy it I get:
```
Specified "redirects" will not automatically work with "output: export". See more info here: 
https://nextjs.org/docs/messages/export-no-custom-routes
```

Feel free to re-open if you think this can be done differently.

joysamaddar:
Reopened the PR in here - https://github.com/DeXter-on-Radix/website/pull/621

